PR #20, opened 2026-04-29
Document CLI usage and fix wrapper imports

mbtiongson1:


mbtiongson1:
web-search instead of tokenize

PR #36, opened 2026-04-29
[reclassify] rag-pipeline — strengthen Named evidence

rico-favor:
### Target Skill
- **ID:** `rag-pipeline`
- **Current Level:** III → **Proposed Level:** III
- **Current Rarity:** uncommon → **Proposed Rarity:** uncommon

### Justification

`rag-pipeline` is already classified at Level III, but its evidence entry was a thin internal seed note. This PR keeps the classification unchanged and strengthens the public evidence supporting it with a peer-reviewed RAG evaluation demo and a reproducible open-source implementation.

The new notes also document concrete failure modes relevant to RAG pipelines:
- retrieval can miss relevant chunks;
- ambiguous queries can retrieve misleading context;
- generated answers can drift from retrieved evidence;
- retrieval quality and answer faithfulness need separate evaluation dimensions.

### New Evidence
| Class | Source | Evaluator | Date | Notes |
|---|---|---|---|---|
| A | https://aclanthology.org/2024.eacl-demo.16/ | rico-favor | 2026-04-29 | RAGAS peer-reviewed demo paper defines reproducible metrics for retrieval-augmented generation, including context relevance, faithfulness, and answer relevance. |
| B | https://github.com/vibrantlabsai/ragas | rico-favor | 2026-04-29 | RAGAS open-source implementation provides a reproducible evaluation toolkit for RAG pipelines and documents RAG evaluation workflows. |

### Existing PRs / related work

Searched open/closed PRs for related RAG evidence work. No duplicate RAG evidence PR was found. This directly addresses #10.

### Validation

```bash
python3 scripts/validate.py
python3 scripts/generateProjections.py
python3 scripts/validate.py
```

### Checklist

**Contributor:**
- [x] Justification references concrete evidence, not opinion.
- [x] New evidence meets the threshold for the target level.
- [x] Rarity is unchanged.
- [x] I have run `python scripts/validate.py` locally and it passes.

Closes #10

PR #35, opened 2026-04-29
[composite] function-calling — schema-driven tool invocation

rico-favor:
### Skill Details
- **ID:** `function-calling`
- **Name:** Function Calling
- **Level:** III
- **Prerequisites:** `structured-output`, `api-call`, `tool-select`
- **Description:** Selects an external callable, emits schema-valid arguments, invokes the function or API, and integrates the returned result into the agent loop.
- **Fusion Condition:** Requires an available function or tool catalog with machine-readable argument schemas and result contracts.

### Evidence
| Class | Source | Evaluator | Date | Notes |
|---|---|---|---|---|
| A | https://arxiv.org/abs/2307.16789 | rico-favor | 2026-04-29 | ToolLLM / ToolBench demonstrates LLM tool invocation over 16,000+ real-world APIs with documented evaluation tasks and failure analysis. |
| A | https://arxiv.org/abs/2305.15334 | rico-favor | 2026-04-29 | Gorilla / APIBench evaluates model generation of API calls across large API collections, including endpoint and argument selection. |

### Rationale

The issue proposed `function-calling` as an atomic skill, but the current graph already has the primitives that make it work: `structured-output`, `api-call`, and `tool-select`. This PR models function calling as the emergent capability of combining those primitives instead of duplicating `tool-use` or `api-call`.

A model that only has `structured-output` can emit valid JSON but not necessarily choose or invoke a tool. A model that only has `api-call` can call APIs but not necessarily choose from a tool catalog or produce schema-valid function arguments. A model that only has `tool-select` can choose a tool but not necessarily execute a complete typed invocation loop. The composite captures the full function-calling loop.

### Existing PRs / related work

Searched open/closed PRs for related function-calling / tool-use work. Closest related merged work is #33, which added or updated adjacent skills including `tool-use`, `api-call`, and `structured-output`; this PR builds on those rather than adding a duplicate atomic primitive.

### Checklist

**Contributor:**
- [x] Skill ID uses `kebab-case` with no vendor references.
- [x] Description describes an emergent capability, not just "all parents at once."
- [x] At least 2 prerequisites listed, all referencing existing skill IDs in `gaia.json`.
- [x] Evidence meets the minimum threshold for the claimed level (Level III = Class B+).
- [x] Fusion condition is specified if applicable.
- [x] I have run `python scripts/validate.py` locally and it passes.
- [x] Generated projections were regenerated with `python3 scripts/generateProjections.py`.

Closes #9

PR #38, opened 2026-04-29
fix(ci): use pull_request_target so route-review can label fork PRs

mbtiongson1:
## Summary
- Fixes `route-review` CI failure on PRs #35 and #36 (both from fork branches)
- Root cause: `pull_request` events from forks get a read-only `GITHUB_TOKEN`; `gh pr edit --add-label` fails with `GraphQL: Resource not accessible by integration (addLabelsToLabelable)`
- Fix: switch trigger to `pull_request_target`, which runs with the base-repo token and has full write access regardless of whether the PR is from a fork

## Security
`pull_request_target` can be dangerous if you run untrusted PR code under elevated permissions. To stay safe:
- Removed `actions/checkout` + `python scripts/validate.py` from the triage jobs (schema/DAG validation is already handled by `validate.yml` using `pull_request`)
- Replaced the `git diff` legendary-proposal check with `gh api` calls (reads file contents via REST, no code execution)
- No PR code is ever executed in this workflow

## Test plan
- [ ] Merge this PR to main
- [ ] Push a new commit to PR #35 or #36 (or re-run the workflow) — `route-review` should pass and add the `human-proposal` + `needs-review` labels
- [ ] Verify `validate.yml` still independently catches schema/DAG errors on PRs touching `graph/**`

### DIFF
--- a/.github/workflows/auto-triage.yml
+++ b/.github/workflows/auto-triage.yml
@@ -1,7 +1,9 @@
 name: "Auto-Triage"
 
+# pull_request_target runs with the base-repo token (full write permissions)
+# even for fork PRs. We never checkout or execute PR code here, so this is safe.
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize]
     paths:
       - "graph/**"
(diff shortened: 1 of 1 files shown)


mbtiongson1:
LGTM

PR #279, opened 2026-05-18
fix(mcp): report package version from server metadata

### LINKED ISSUE #214: mcp: server reports hardcoded version 0.1.0 instead of package.json version
The MCP server advertises `version: "0.1.0"` in its `initialize` response regardless of the actual package version.

**File:** `packages/mcp/src/index.ts` (serverInfo block)
**Impact:** MCP clients that check server version see a stale/wrong value. Will cause confusion as the package version bumps.

**Fix:** Read version dynamically from `package.json` at startup, e.g.:
```ts
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
const { version } = require('../../package.json');
```

rico-favor:
## Summary
- Reads the MCP server version from `packages/mcp/package.json` instead of hardcoding `0.1.0`
- Handles both source (`src/`) and compiled (`dist/src/`) runtime paths
- Adds a protocol test asserting initialize metadata matches the package version
- Updates the MCP package lock root version to match `package.json`

## Validation
- `cd packages/mcp && npm run build`
- `cd packages/mcp && npm test`
- `node -e "import('./dist/src/index.js').then(m=>{m.createServer(); console.log('compiled import ok')})"`

Closes #214

### DIFF
--- a/packages/mcp/package-lock.json
+++ b/packages/mcp/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@gaia-registry/mcp-server",
-  "version": "3.0.1",
+  "version": "3.11.15",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@gaia-registry/mcp-server",
-      "version": "3.0.1",
+      "version": "3.11.15",
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.0",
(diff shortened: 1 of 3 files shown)


mbtiongson1 (approved):
Approve. The fix is correct, the path-probing logic is sound given the project's TS layout, and the test properly validates the behavior. The issues above are all minor and pre-existing or low-risk.

PR #280, opened 2026-05-18
fix(cli): normalize leading slash skill IDs

### LINKED ISSUE #180: gaia skills info/install: leading slash in skill ID causes crash and double-slash error
## Summary

Two related bugs when a user passes a skill ID with a leading slash (e.g. `/huggingface/hf-cli` instead of `huggingface/hf-cli`):

### Bug 1 — `gaia skills info` double-slash error

**Repro:**
```
python -m gaia_cli skills info /huggingface/hf-cli
```
**Expected:** skill card for `huggingface/hf-cli`  
**Actual:** `Skill '//huggingface/hf-cli' not found.`

The error message at `main.py:981` prepends `/` to `q` unconditionally:
```python
print(f"Skill '/{q}' not found.", file=sys.stderr)
```
But when the user already included a leading slash (`q = "/huggingface/hf-cli"`), the comparison `item.get("id") == q` fails (registry IDs have no leading slash) and the double-slash message is shown.

### Bug 2 — `gaia skills install` crashes with `FileNotFoundError`

**Repro:**
```
python -m gaia_cli skills install /huggingface/hf-cli
```
**Expected:** skill installed  
**Actual:** `FileNotFoundError: [WinError 3] The system cannot find the path specified`

Root cause (`install.py:114-136`):

1. `resolve_named_skill_reference("/huggingface/hf-cli", ...)` calls `find_named_skill_source`, which does `"/huggingface/hf-cli".split("/", 1)` → `contributor=""`, `skill_name="huggingface/hf-cli"`. Python's `os.path.join(dir, "", "huggingface/hf-cli.md")` silently resolves to the real file path, so resolution "succeeds" but returns the un-normalized skill ID `/huggingface/hf-cli`.

2. Back in `install_skill`, the skill_id is still `/huggingface/hf-cli`. The same bad split produces `contributor=""` and `skill_name="huggingface/hf-cli"` (contains a `/`). `os.makedirs(global_dir)` creates the wrong directory, leaving the destination parent for `shutil.copy2` non-existent → crash.

## Fix

Strip leading `/` from skill IDs at the CLI entry point (or inside `resolve_named_skill_reference` / `install_skill`). See `main.py:978` and `install.py:84`.

A one-liner like `skill_ref = skill_ref.lstrip("/")` at the top of `resolve_named_skill_reference` would fix Bug 2. `main.py:978` needs `q = args.skill_id.lstrip("/")` to fix Bug 1.

## Environment

- Platform: Windows 11  
- Python: 3.13  
- Branch tested: `docs/update-2026-05-08`

rico-favor:
## Summary
- Normalizes slash-prefixed named skill IDs in `gaia skills info`
- Normalizes install references before resolving named skill files, catalog refs, or bare slugs
- Adds coverage for installing `/contributor/skill` as the canonical `contributor/skill` ID

## Validation
- `PYTHONPATH=src python3 -m py_compile src/gaia_cli/install.py src/gaia_cli/main.py tests/test_install.py`
- `PYTHONPATH=src python3 -m gaia_cli.main --registry . skills info /huggingface/hf-cli`
- `GAIA_HOME=... PYTHONPATH=... python3 -m gaia_cli.main --registry ... skills install /huggingface/hf-cli`

Note: `pytest` is not installed in this Termux environment, so I verified syntax plus the affected CLI flows directly.

Closes #180

### DIFF
--- a/src/gaia_cli/install.py
+++ b/src/gaia_cli/install.py
@@ -81,6 +81,7 @@ def resolve_named_skill_reference(skill_ref, registry_path):
     - exact catalogRef frontmatter slug
     - unique bare skill-name slug
     """
+    skill_ref = skill_ref.lstrip("/")
     source = find_named_skill_source(skill_ref, registry_path)
     if source:
         return skill_ref, source
--- a/src/gaia_cli/main.py
+++ b/src/gaia_cli/main.py
@@ -1156,7 +1156,7 @@ def skills_command(args):
         ctx = None
 
     if verb == "info":
-        q = args.skill_id
+        q = args.skill_id.lstrip("/")
         match = next((item for item in items if item.get("id") == q), None)
         if not match:
             print(f"Skill '/{q}' not found.", file=sys.stderr)
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -12,6 +12,7 @@ from gaia_cli.install import (
     find_named_skill_source,
     install_skill,
     load_manifest,
+    resolve_named_skill_reference,
     save_manifest,
     uninstall_skill,
     list_installed,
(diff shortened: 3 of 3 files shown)


mbtiongson1 (approved):
The install side is correct and well-tested. The uninstall side has the same bug and should be fixed. Adding a test for uninstall_skill("/...") and ideally one for skills info /... would  make this a complete fix.

I'm approving this one, will work on the uninstall and tests in a different issue. Thanks!

Issue #277, opened 2026-05-18
garrytan/cso has dependencies


mbtiongson1:
Some skills like Garry Tan's /gstack and other skills under it depends on the installation of the whole suite.

This means that standalone skills cannot be installed via npx or even gaia install.

Some corrections are needed:

- Do we introduce a "suite" installation? much like "/superpowers"
- how different will this be from an "agent" installation?
- how will this look like in the gaia tree? is this searchable, or are they "fused" like /grill-me-docs (from /ubiquitous-language and /grill-me)

PR #291, opened 2026-05-18
infra(skills): add /gaia-fuse-full-suite workflow skill

mbtiongson1:
## Summary

Adds `.claude/skills/gaia-fuse-full-suite/skill.md` — a Claude Code slash-command that automates fusing a named contributor's full skill suite into a single ultimate skill node.

Motivated by the manual work done in PR #290 (obra/superpowers), this skill encodes that process as a reusable, step-by-step workflow.

## What the skill does

1. **Collect components** — reads `registry/named/<contributor>/` to find all named skills and their `genericSkillRef` IDs
2. **Verify nodes exist** — aborts if any component node is missing from `registry/nodes/`
3. **Guard against re-fusion** — aborts if an existing ultimate already has a `fuse` timeline entry
4. **Research evidence** — fetches the source URL and applies a defined 5★ vs 4★ decision rule (10k+ stars + multi-platform → 5★; otherwise 4★)
5. **Create/update ultimate node** — writes `registry/nodes/ultimate/<id>.json` with fuse timeline event
6. **Back-link components** — appends the ultimate ID to each component node's `derivatives`
7. **Update `registry/gaia.json`** — adds skill entry + prerequisite edges, deduplicating
8. **Regenerate + validate** — runs `generateNamedIndex.py` and `validate.py` (all 10 checks must pass before PR)
9. **Commit + open PR** on `review/meta/<ultimate-id>-fusion`

## Test plan

- [ ] Invoke `/gaia-fuse-full-suite` on a contributor with 3+ named skills and verify it produces a clean PR
- [ ] Verify abort when the ultimate already exists with a fuse timeline entry
- [ ] Verify 4★ path when source URL has < 10k stars


mbtiongson1:
#277 will now be solved--these "full-suite" skills will be the ultimates. None in the schema will be "forcing" the user to install the whole suite--but maybe in the design, this will be incorporated. Right now, this is the whole reason the "ultimates" are here--to fuse things into the whole suite of skills that users use today.

PR #2, opened 2026-04-28
[multi] Add 30 popular AI agent skills at Level IV with real evidence

mbtiongson1:
## Summary

- Adds **30 new skills** (17 atomic, 10 composite, 3 legendary) to `graph/gaia.json`, all at **Level IV (Proficient)** with at least one Class A or B evidence entry
- Evidence sources are real peer-reviewed papers (arXiv) and reproducible open-source repos
- Passes all 6 validator checks: schema, DAG, references, prereq counts, evidence thresholds, legendary constraints
- Graph grows from 36 to **66 skills** and 35 to **74 edges**

### New atomic skills (17)
`translate`, `sentimentAnalysis`, `imageCaption`, `speechToText`, `textToSpeech`, `generateSql`, `questionAnswer`, `imageGenerate`, `mathReason`, `logicalInference`, `memoryManage`, `apiCall`, `refactorCode`, `generateTest`, `parsePdf`, `detectAnomaly`, `dataVisualize`

### New composite skills (10)
`conversationalAgent`, `textToSqlPipeline`, `codeReviewPipeline`, `dataAnalysis`, `voiceAgent`, `automatedTesting`, `contentModeration`, `multimodalReasoning`, `translationPipeline`, `documentDigitization`

### New legendary skills (3) — each with 3 Class A/B sources
`fullStackDeveloper`, `autonomousDataScientist`, `realTimeVoiceAssistant`

### Key evidence papers
NLLB (2207.04672), Whisper (2212.04356), BLIP-2 (2301.12597), VALL-E (2301.02111), Spider (1809.08887), SQuAD 2.0 (1806.03822), Stable Diffusion (2112.10752), Minerva (2206.14858), Chain-of-Thought (2201.11903), MemGPT (2310.08560), Gorilla (2305.15334), SWE-bench (2310.06770), Codex/HumanEval (2107.03374), Nougat (2308.13418), DIN-SQL (2304.11015), CodeReviewer (2203.09095), LLaVA (2304.08485), SeamlessM4T (2308.11596), CodeAct (2402.01030), DS-1000 (2210.12641), AudioPaLM (2312.11805), VoiceBox (2305.11206)

## Test plan

- [x] `python3 scripts/validate.py` — all 6 checks pass, zero errors
- [x] 66 total skills, 74 edges, max lineage depth 3
- [ ] Run `python3 scripts/generateProjections.py` after merge to regenerate `skills/`, `registry.md`, `combinations.md`
[...]

PR #13, opened 2026-04-28
docs: add curated SKILL.md source contributions and link in CONTRIBUTING.md

mbtiongson1:
### Motivation

- Provide contributors a practical starting point for discovering community skill definitions authored as `SKILL.md` so proposals can be grounded in existing ecosystem examples.
- Reduce duplicate or low-quality proposals by surfacing where reusable skill artifacts live and how to treat them as evidence candidates.

### Description

- Added a new research page at `docs/skill_source_contributions.md` containing a short selection methodology, a curated list of high-usage GitHub repos that publish `SKILL.md` assets, point-in-time star counts, and contributor guidance.
- Linked the new research page from `CONTRIBUTING.md` by adding a `## Community Skill Source Research` section that points contributors to the document.
- The new doc emphasizes treating star counts as heuristics and reminds contributors to map candidates to existing Gaia IDs and attach reproducible evidence per Gaia's rubric.

### Testing

- Ran `python3 scripts/validate.py` against the canonical graph, and all validation checks passed in this environment.
- The validator reported the JSON schema sub-check was skipped because `jsonschema` is not installed (`Install with: pip install jsonschema`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_69f12896afd88327a3850f8156656802)

Issue #12, opened 2026-04-28
Add a GitHub Pages skill browser (read-only view of registry.md)


mbtiongson1:
## Idea

Right now the only way to browse the skill graph is to read `registry.md` or `tree.md` on GitHub. A GitHub Pages site would give the project a live, linkable URL and make it much easier to share.

## Minimal version

The simplest implementation: enable GitHub Pages on `main` pointing at `docs/` (or root), and add a `docs/index.md` that is just a styled render of `registry.md`.

## Stretch version

A static site (e.g. built with MkDocs, Docusaurus, or plain HTML) with:
- Searchable skill table
- Clickable skill cards showing tier, rank, prerequisites, and evidence
- Auto-regenerated on push via GitHub Actions

## Why this matters

A hosted URL gives journalists, blog posts, and tweets a link to point at instead of a raw GitHub tree. It also makes the project feel "real" to visitors who aren't developers.

## How to start

1. Enable GitHub Pages in repo Settings → Pages → Source: `main` / `docs`
2. Add `docs/index.md` with a header and embed of `registry.md`
3. That's a valid MVP — iterate from there

Happy to review any approach.

rico-favor:
I opened a PR for the MVP docs entrypoint here: https://github.com/mbtiongson1/gaia-skill-tree/pull/34

Scope is intentionally small: a static `docs/index.md` that links to the generated registry/tree/combinations views and contributor docs, without adding a site framework yet.

Issue #10, opened 2026-04-28
Level up RAG Pipeline: Dormant → Named (add Class B evidence)


mbtiongson1:
## Task

`RAG Pipeline` is currently at **III · Named** in the graph but its evidence entry is thin. This issue is to attach real Class B evidence and document known failure modes so it can be confirmed at rank III.

## What Class B evidence looks like

- A paper or benchmark result showing reproducible RAG performance (e.g. RAGAS, BEIR)
- A blog post or repo with a full working implementation + documented edge cases
- A failure mode write-up (e.g. "retrieval fails when query is ambiguous")

## How to contribute

1. Find a solid public evidence source
2. Edit the `rag-pipeline` entry in `graph/gaia.json` — add `evidence` with `class: B`, `url`, and `notes`
3. Run `python3 scripts/validate.py`
4. PR it in — label it `evidence` and the auto-triage workflow handles the rest

See [CONTRIBUTING.md](CONTRIBUTING.md) for evidence class definitions.

rico-favor:
I opened a PR to strengthen the `rag-pipeline` evidence here: https://github.com/mbtiongson1/gaia-skill-tree/pull/36

It keeps the level/rank unchanged, replaces the thin seed evidence with public RAGAS evidence, and documents concrete RAG failure modes in the evidence notes.

PR #325, opened 2026-05-20
fix(cli): gaia dev suite + push scanner fixes (Issues #322, #337, #338)

### LINKED ISSUE #322: meta review operations updates: gaia cli should be used when merging or deduplicating, etc.
Meta reviews are too token heavy and LLMs usually can't keep up with much data--there should be a CLI to directly deduplicate skills or just list generic skill references and named skill references of a certain skill. Skill ids should also be added for unique ID numbered references for a stateful canon tree.

This way, meta audit will use automated scripts that use the CLI tool instead of just relying on LLM logic. Will add functionality, and will be good for automated meta audit runs in the future.

#Suggested workflow:
1. gaia list
* --generic to get the list of all generic skill names
* --named to get the list of all named skill names
* --description will also pass the description
* --xxxx will pass any schema 
2. gaia merge <target skill1> <skill2> <skill3>... [2+ arguments]
* adding --named will merge named skills into a bucket
* will deduplicate skill with target to the target skill ref
* will add schemas together, and collapse others. Ex: Evidences will be stacked, while skill-name will collapse. Named skills will be added to buckets, with only one origin named. maybe pass through --origin to name the origin, or just automatically reference the earliest one.
3. update meta timelines
4. optional: gaia split <> <>
* will split a generic skill ref skill into two or three with a suggested name
5. gaia add <skill-name> <schema>
* --generic or --named can be passed
* updates schema automatically

This way, json files doesn't have to be manually edited by the LLM--will be less token heavy (hopefully)l.

mbtiongson1:
## Summary

Closes #322 — implements `gaia dev` subcommand suite for registry maintenance.
Fixes Issues #337 and #338 — aggressive push scanning + `namedSuggestions` schema violation.
Verifies Issue #333 — heavyweight ML deps are lazy-imported behind the `[embeddings]` extra.

### `gaia dev` subcommand suite (Issue #322)
- `dev list`, `dev add`, `dev merge`, `dev split`, `dev rename`, `dev evidence`, `dev audit`, `dev calibrate`
- All meta operations go through CLI (no raw file edits) to guarantee schema integrity + timeline logging

### 14 audit bugs closed
| Bug | Fix |
|-----|-----|
| #1 | Tests and CLAUDE.md updated to use `gaia dev <cmd>` surface |
| #2 | `_run_docs_build(args.registry)` forwards `--registry` to subprocess |
| #3 | `dev add` rejects descriptions shorter than 10 chars |
| #4 | Merge deduplicates derivatives/prerequisites after source replacement |
| #5 | Calibrate validates level against allowed enum |
| #6 | Rename checks for collision before writing new file |
| #7 | Split uses `copy.deepcopy`; checks for collision per target |
| #8 | Audit handles malformed `level` fields defensively |
| #9 | Schema: added `rename` to `timelineEvent.action` enum (both schema copies) |
| #10/#11 | Named-skill lookups resolve by frontmatter `id`, not body substring |
| #12 | Dead top-level alias dispatch removed from `main.py` |
| #13 | Downstream of #2 — tests no longer touch real registry |
| #14 | Named `dev add` writes frontmatter via `yaml.dump`, not f-string |

### Push scanner fix (Issues #337/#338)
- Scanner regex now requires `/` prefix (e.g. `/web-search`) — prevents every common word matching as a skill
- Removed `namedSuggestions` from push batch schema (invalid field)
- Leading-slash stripping added in `build_skill_batch`, `scan_command`, `hook.py`, and `pathEngine.py` so resolved skill IDs remain clean

### Issue #333 verification
- No eager top-level imports of `sentence_transformers`, `torch`, or `numpy`
- `embed_command` and `search_command` both guard `ImportError` gracefully
- `yaml` import in `dev.py` made lazy (avoids eager-import regression in `--no-deps` installs)

## Test results
- `tests/test_meta_ops.py` — 6/6 pass (includes new short-description regression test)
- Full suite — 373 pass, 3 pre-existing failures (ANSI color in `--help` output, present on `main` before this PR)

## Pre-existing failures (not introduced here)
- `test_bare_skills_command_prints_skills_help`
- `test_python_module_help_runs_with_gaia_prog_name`
- `test_wheel_install_smoke_tests_console_script`

All three fail on `origin/main` for the same reason (ANSI escape codes in argparse help output).


mbtiongson1:
## 🐛 Bug #1: `tests/test_meta_ops.py` — all 5 tests fail (CLI surface mismatch)

The new test file added in this PR exercises top-level subcommands that don't exist. They were moved under `gaia dev`.

**Result:** `pytest tests/test_meta_ops.py` fails 5/5 with `SystemExit: 2`.

```
gaia: error: argument {...}: invalid choice: 'evidence' (choose from ..., dev, validate, test, skills, ...)
gaia: error: argument {...}: invalid choice: 'merge'
gaia: error: argument {...}: invalid choice: 'add'
gaia: error: argument {...}: invalid choice: 'list'
```

The tests invoke:
- `gaia ... list --generic` → must be `gaia dev list --generic`
- `gaia ... list --named` → must be `gaia dev list --named`
- `gaia ... add Skill C ...` → must be `gaia dev add ...`
- `gaia ... merge skill-a skill-b` → must be `gaia dev merge ...`
- `gaia ... evidence skill-a ...` → must be `gaia dev evidence ...`

**Also a docs/impl mismatch:** `CLAUDE.md` still documents these as top-level (`gaia list --generic --named`, `gaia add ...`, `gaia merge ...`, `gaia evidence ...`).

**Fix:** Either (a) update `tests/test_meta_ops.py` argv to insert `"dev"` between `--registry <path>` and the subcommand, plus update `CLAUDE.md`, **or** (b) also register these as top-level aliases so the documented invocation works.

mbtiongson1:
## 🐛 Bug #2: `gaia dev` subcommands don't propagate `--registry` to `docs build` subprocess

Every dev write command ends with:

```python
subprocess.run([sys.executable, "-m", "gaia_cli", "docs", "build"], check=True)
```

`--registry` from the parent invocation is dropped. When a user runs `gaia --registry /path/to/checkout dev add ...`, the subprocess re-resolves a different registry and errors:

```
gaia: error: `gaia docs` needs a writable registry. Either:
  • Run `gaia init` from your gaia-skill-tree clone …
  • Pass --registry PATH explicitly.
Created generic skill: /tmp/.../registry/nodes/basic/test-skill.json
Regenerating registry and documentation...
CalledProcessError: returned non-zero exit status 2.
```

Affects: `meta_add_command`, `meta_merge_command`, `meta_split_command`, `meta_rename_command`, `meta_calibrate_command`, `meta_evidence_command` (all in `src/gaia_cli/commands/dev.py`).

**Fix:** Forward `--registry` when present:
```python
cmd = [sys.executable, "-m", "gaia_cli"]
if args.registry:
    cmd += ["--registry", str(args.registry)]
cmd += ["docs", "build"]
subprocess.run(cmd, check=True)
```

**Bonus side-effect:** the skill file is written *before* `docs build` runs, so a failed regeneration leaves an orphan JSON node on disk with no rollback.

---

## 🐛 Bug #3: `gaia dev add` produces schema-invalid JSON (regression of Issue 322 fix #3)

When `--description` is omitted, the generated node has `"description": null`. When a short description is supplied, the field violates `minLength: 10`.

Reproduction:
```bash
gaia --registry /tmp/sb dev add TestSkill --id test-skill
cat /tmp/sb/registry/nodes/basic/test-skill.json
# {
#   "id": "test-skill",
#   ...
#   "description": null,   ← schema requires string, minLength 10
#   ...
# }
```

Root cause in `meta_add_command`:
```python
"description": getattr(args, "description", ""),
```
`getattr` returns the default *only* when the attribute is missing. argparse always sets `args.description`, defaulting to `None` — so the `""` fallback never fires.

**Fix:** `(getattr(args, "description", None) or "").strip()` and either reject when shorter than 10 chars or pad with a placeholder + warn. Also: `--description` is not a required argument; either make it required or validate before write.

The Issue 322 comment claims this was fixed; the test in `tests/test_meta_ops.py::test_meta_add_generic` only passes a description long enough to satisfy the schema, so the regression isn't covered.

mbtiongson1:
## 🐛 Bug #4: `gaia dev merge` silently fails to dedup parent's `derivatives` when target is already present (regression of Issue 322 fix #2)

Issue 322 specifically called out "merge fails to clean up derivatives in parent nodes" and the PR claims this is fixed. It isn't — for the common case.

### Reproduction
Parent skill `P` has both `child-a` and `child-b` as derivatives. Merge `child-b` into `child-a`:

```bash
gaia --registry /tmp/m dev merge child-a child-b
```

**Expected:** `P.derivatives == ["child-a"]`
**Actual:** `P.derivatives == ["child-a", "child-b"]` (stale `child-b` reference remains; will fail reference-integrity validation)

### Root cause (`src/gaia_cli/commands/dev.py`, lines 294-303)

```python
if "derivatives" in data:
    new_derivatives = []
    for dr in data["derivatives"]:
        if dr in sources:
            if target_id not in new_derivatives and target_id != data["id"]:
                new_derivatives.append(target_id)
                changed = True            # ← only set when target gets appended
        else:
            new_derivatives.append(dr)
    data["derivatives"] = new_derivatives
```

Walk-through for parent with `["child-a", "child-b"]`, target=`child-a`, sources=`["child-b"]`:
1. `dr="child-a"` → not in sources → appended → `new=["child-a"]`, `changed=False`
2. `dr="child-b"` → in sources → check `target_id in new_derivatives` → **True** → skip the append, **`changed` stays False**

`new_derivatives` is `["child-a"]` but because `changed == False` the file write at L305-308 never runs, so the stale `child-b` reference is left on disk.

Same shape of bug exists for `prerequisites` at L283-292 — a parent whose prerequisites list already contains the target won't get the source dropped.

### Fix
Compare lists before/after, not just on the "needed to add target" branch:

```python
if data.get("derivatives"):
    new_derivatives = []
    for dr in data["derivatives"]:
        if dr in sources:
            if target_id not in new_derivatives and target_id != data["id"]:
                new_derivatives.append(target_id)
        else:
            new_derivatives.append(dr)
    if new_derivatives != data["derivatives"]:
        data["derivatives"] = new_derivatives
        changed = True
```

(Mirror the same fix for `prerequisites`.)

### Additional consequence — partial-state on failure
Even before this bug bites, `gaia dev merge` deletes the source JSON files (L269-271) **before** invoking the failing `docs build` subprocess from Bug #2. The source skill is gone, parents still point at it, and there is no rollback path. Wrapping the destructive section in a try/except that restores deleted files on failure would harden this.

mbtiongson1:
## 🐛 Bug #5: `gaia dev calibrate` accepts levels outside the schema enum

The validator only checks "ends with ★ and starts with a digit." `99★`, `7★`, etc. are accepted and written to disk even though `skill.schema.json` restricts `level` to `enum: [0★…6★]`.

```bash
gaia --registry /tmp/sb dev calibrate test 99★
# Successfully calibrated 'test' to 99★.
# → "level": "99★", "action": "rank_up", "details": "Calibrated level from 2★ to 99★"
```

The invalid value only surfaces on the next `gaia validate`, after a timeline event has been recorded and a downstream `docs build` has propagated it.

**Fix:** Either parse the allowed enum from the schema and validate against it, or hardcode `{0,1,2,3,4,5,6}`.

Bonus: the rank-up/demote heuristic at L438-441 uses `int(level[0])` which truncates two-digit levels to their first character (`"10★"[0] == "1"`). If you ever raise the cap, calibrating from `5★` to `10★` will silently log as `demote`.

---

## 🐛 Bug #6: `gaia dev rename` silently overwrites an existing skill (data-loss)

There is no collision check on the destination ID.

```bash
# pre: foo.json and bar.json both exist
gaia --registry /tmp/ren dev rename foo bar
# Renamed .../foo.json to .../bar.json
# Successfully renamed 'foo' to 'bar'.
ls .../basic/   # → only bar.json (the original bar is destroyed)
cat .../bar.json
# { "id": "bar", "name": "Foo", ... }   ← Foo's data with bar's id
```

Root cause (`src/gaia_cli/commands/dev.py`, L350-358):
```python
new_file = old_file.parent / f"{new_id}.json"
...
with open(new_file, "w", encoding="utf-8") as f:
    json.dump(skill_data, f, ...)
old_file.unlink()
```

`open(..., "w")` overwrites unconditionally. The original `bar` skill is lost with no warning, no backup, no timeline event.

**Fix:** Before writing the new file, search the registry for any node where `id == new_id` (mirroring the existing search for `old_id`). If one exists, abort with a clear error. Same fix needed in `meta_split_command` (L539-543) — `target_id` collisions when creating split targets are also unchecked.

mbtiongson1:
## 🐛 Bug #7: `gaia dev split` shares list references and duplicates evidence

`meta_split_command` at L532-538:
```python
for target_id in targets:
    target_data = source_data.copy()      # shallow copy
    target_data["id"] = target_id
    target_data["name"] = target_id.replace("-", " ").title()
    ...
```

`dict.copy()` is shallow, so every split target shares the **same list object** for `evidence`, `prerequisites`, `derivatives`, `knownAgents`, and `timeline`. Demonstrated:

```bash
gaia --registry /tmp/sp dev split origin split-a split-b
python3 -c "import json; a=json.load(open('split-a.json')); b=json.load(open('split-b.json')); print(a['evidence']==b['evidence'])"
# → True   (same reference)
```

Two side-effects:
1. **Mutation aliasing** — after split, mutating `split-a.evidence` mutates `split-b.evidence` in-memory before the next write. Subtle, only bites later workflows.
2. **Evidence duplication** — a single source evidence URL now appears under *two* canonical skills. Validators that look for evidence uniqueness or contributor-attribution will be wrong.

**Fix:** Use `copy.deepcopy(source_data)`. Also: decide whether the original evidence should be retained on *one* target (the canonical one), or stripped from all and require fresh evidence per split.

**Also:** No collision check — if a target ID matches an existing skill, the existing file is overwritten (same family of issue as Bug #6).

---

## 🐛 Bug #8: `gaia dev audit` crashes on any skill with an empty `level` string

```python
level_str = data.get("level", "0★")
level = int(level_str[0])           # ← IndexError when level_str == ""
```

`data.get("level", "0★")` only returns the default when the key is *missing*; an explicit `"level": ""` (or `null`) falls through to `level_str[0]` → `IndexError: string index out of range`. The audit aborts entirely; no other issues in the registry are reported.

```bash
gaia --registry /tmp/au dev audit
# EXCEPTION: IndexError string index out of range
```

**Fix:**
```python
level_str = data.get("level") or "0★"
try:
    level = int(level_str[0])
except (ValueError, IndexError):
    issues.append(f"[P0] {skill_id}: Malformed level {level_str!r}")
    continue
```

Apply the same hardening to `meta_calibrate_command`'s `int(old_level[0])` (L439).

---

## 🐛 Bug #9: `gaia dev rename` writes a schema-invalid timeline event

`meta_rename_command` calls:
```python
append_skill_event(new_id, "rename", _get_contributor(), ...)
```

But `skill.schema.json`'s `timelineEvent.action` enum is:
```
rank_up, fuse, propose, name, demote, ascend, rejected, merge, split, add, evidence_added, type_change
```

There is no `rename`. Every rename produces a node that fails `gaia validate`.

**Fix:** Either add `rename` to the enum in `registry/schema/skill.schema.json` *and* `src/gaia_cli/data/registry/schema/skill.schema.json` (both copies — see Bug #11 below), or reuse `type_change`/another existing action and put the old/new id pair in `details`.

mbtiongson1:
## 🐛 Bug #10: `gaia dev merge` for named skills matches `id:` substrings in **body text**, corrupting unrelated files

The named-skill merge finds the target file by raw substring match on the *entire* file content (frontmatter and body):

```python
for p in named_dir.glob("**/*.md"):
    with open(p, "r", encoding="utf-8") as f:
        content = f.read()
        if f'id: "{target_id}"' in content or f"id: '{target_id}'" in content or f"id: {target_id}" in content:
            target_file = p
            break
```

Any markdown file that mentions the target id in body prose (a README, a cross-reference, an example) is treated as the target. Glob iteration order is alphabetic, so the first matching file wins.

### Reproduction
```
named/alice/decoy.md   ← body says: This file references id: "alice/target" in body text.
named/alice/src.md
named/alice/target.md  ← actual target
```
```bash
gaia --registry /tmp/nm dev merge alice/target alice/src
# "Successfully merged skills into 'alice/target'."
```

**Actual result:**
- `decoy.md` has `src.md`'s body appended **after its own original body** under `--- Merged from alice/src ---`
- `target.md` is **untouched**
- `src.md` is deleted
- The frontmatter in `decoy.md` has been rewritten through `yaml.dump`, stripping the original quoting (`id: alice/decoy` instead of `id: "alice/decoy"`) — collateral churn that will diff-bomb the next git operation.

This is destructive data corruption. The same shape exists for the source-file lookup at L165-174.

### Fix
Resolve the file by parsing YAML frontmatter and checking `frontmatter["id"] == target_id`, not by string-searching the whole file. The lookup helper used in `_update_named_skill_ref` should be replaced by a `frontmatter['id']` equality check across all named files. Pseudocode:

```python
def _find_named_file(named_dir: Path, skill_id: str) -> Path | None:
    for p in named_dir.glob("**/*.md"):
        meta, _ = parse_md(p)
        if meta.get("id") == skill_id:
            return p
    return None
```

Use this in both target and source lookups in `meta_merge_command`.

mbtiongson1:
## 🐛 Bug #11: `_update_named_skill_ref` mutates body text outside frontmatter and re-quotes values

The helper uses raw `str.replace()` over the whole file:
```python
if old_line in content:
    content = content.replace(old_line, new_line)
```
This rewrites every occurrence — frontmatter and body. With prose like `see also genericSkillRef: oldref for context` in a markdown body, a `gaia dev rename oldref newref` mutates that prose to `see also genericSkillRef: "newref" for context`. The output also unconditionally re-quotes the value (`"{new_ref}"`) regardless of the original style, churning files unnecessarily.

**Fix:** Parse the frontmatter (e.g. `parse_md`), update `meta["genericSkillRef"] = new_ref`, and write back. Don't touch body text.

---

## 🐛 Bug #12: Dead-code dispatch path for top-level meta commands

`src/gaia_cli/main.py` L1609:
```python
elif args.command == 'dev' or args.command in ('list', 'merge', 'split', 'add', 'rename', 'calibrate', 'evidence'):
```

The parser only registers these names under `dev` (and `list` under `skills`), never at top level. The `or args.command in (...)` branch is unreachable — argparse rejects them before this dispatch runs. Either (a) remove the dead clause, or (b) register top-level aliases so the documented invocation in `CLAUDE.md` actually works.

Also `audit` is missing from that tuple, so even if top-level aliases were added, `gaia audit` would still fall through to `parser.print_help()` while `gaia merge` etc would dispatch correctly. Inconsistent.

---

## 🐛 Bug #13: New `tests/test_meta_ops.py` pollutes the real registry through Bug #2 spillover

Each meta-op test calls `gaia ... merge|add|evidence ...`. Those commands invoke a subprocess `gaia docs build` without forwarding `--registry`. Run from the repo, the subprocess auto-resolves `args.registry` to the **repo's own** `.gaia/config.toml` and rebuilds docs against the *real* registry — not the `tmp_path` fixture.

Side-effects:
- Tests are not isolated. They touch `registry/` and the generated `docs/` in the working tree.
- A CI run can produce diffs in tracked files (or fail because the working tree is read-only).
- The tests themselves don't assert on subprocess output, so the pollution is invisible.

Fix once Bug #2 is fixed (forward `--registry` to the subprocess). Until then, the test suite should mock `subprocess.run` or set `GAIA_REGISTRY` so the test process can't reach the real registry.

mbtiongson1:
## 🐛 Bug #14: `gaia dev add --named` writes broken YAML when the name/description contains `"`

The named-skill `add` branch builds frontmatter with raw f-strings — no escaping:

```python
content = f"""---
id: "{contributor}/{skill_id}"
name: "{skill_name}"
...
description: "{getattr(args, 'description', '')}"
...
---
"""
```

Reproduction:
```bash
gaia --registry /tmp/na dev add 'Quote "Test" Name' --id quoted --named \
  --contributor tester --generic-ref some-ref --description 'Has a "quote" inside'
```

The file is written, but its frontmatter is unparseable YAML:
```yaml
---
id: "tester/quoted"
name: "Quote "Test" Name"            ← unterminated/extra quotes
description: "Has a "quote" inside"  ← same problem
...
---
```

```
yaml.safe_load: while parsing a block mapping … expected <block end>, but found '<scalar>'
```

The next `gaia dev merge` or anything else that calls `parse_md` on this file blows up. The named-skill index regenerator (via `docs build`) may also fail silently and drop the skill from `named-skills.json`.

**Fix:** Use `yaml.dump` to construct the frontmatter (the same approach the merge command already uses), or `json.dumps(skill_name)` for the value to get correct escaping. Apply to both name and description fields.

mbtiongson1:
## 🔎 Audit summary — 14 bugs filed (Issue #322 follow-up)

Reviewed against the bug list in [Issue #322 comment](https://github.com/mbtiongson1/gaia-skill-tree/issues/322#issuecomment-4493250752). The PR description states all 5 Issue 322 bugs are fixed; verification shows **regressions on #2 and #3**, plus several new issues introduced by the dev-suite.

### Filed in this thread
| # | Severity | Title |
|---|---|---|
| 1 | P0 | `tests/test_meta_ops.py` — 5/5 tests fail (top-level vs `dev` mismatch) |
| 2 | P0 | dev commands drop `--registry` when invoking `docs build` subprocess |
| 3 | P1 | `gaia dev add` writes `description: null` / sub-10-char → schema-invalid (Issue 322 #3 not actually fixed) |
| 4 | P0 | `gaia dev merge` skips updating parent's `derivatives` when target is already present (Issue 322 #2 not actually fixed) |
| 5 | P2 | `gaia dev calibrate` accepts levels outside enum (e.g. `99★`); `int(level[0])` mishandles 2-digit |
| 6 | P0 | `gaia dev rename` silently **overwrites** existing skill at the new id (data loss) |
| 7 | P1 | `gaia dev split` uses shallow `dict.copy()` — list aliasing + evidence duplicated to all targets |
| 8 | P1 | `gaia dev audit` crashes (`IndexError`) on empty `level` string |
| 9 | P0 | `gaia dev rename` writes schema-invalid `action: "rename"` timeline event |
| 10 | P0 | `gaia dev merge` named: body-text substring match → merges into **wrong file**, data corruption |
| 11 | P2 | `_update_named_skill_ref` edits body prose, not just frontmatter; re-quotes values |
| 12 | P3 | Dead-code dispatch path in `main.py` for top-level meta names (and `audit` missing from the tuple) |
| 13 | P1 | `tests/test_meta_ops.py` taints the real registry through Bug #2 spillover |
| 14 | P1 | `gaia dev add --named` writes broken YAML when name/description contain `"` |

### Issue 322 bug-list status (per PR claim)
- **#1 PEP 668**: ✅ Fixed in `update_command` (verified at `src/gaia_cli/main.py:1260-1263`).
- **#2 merge derivatives cleanup**: ❌ **regression** — see Bug #4 above.
- **#3 `gaia add` schema compliance**: ⚠️ partial — `knownAgents` is initialized but `description: null` still produced (Bug #3).
- **#4 `$()` substitution block**: out of scope for this PR (tooling, not CLI).
- **#5 `gaia list` metadata**: ✅ Implemented as `gaia dev list --level/--evidence/--contributor/--extra`.

### Quick wins
- Forwarding `--registry` to the `docs build` subprocess (Bug #2) is a one-line fix and would also unblock Bug #13.
- Adding collision checks (Bugs #6, #7) is small and prevents silent data loss.
- Replacing the substring-based file resolution (Bugs #10, #11) with YAML-frontmatter parsing closes the worst data-corruption surface.

Happy to follow up with patches if useful.

PR #348, opened 2026-05-20
experiment(ml-graph-viz): ML layout sampler for skill graph

mbtiongson1:
## Summary

- Adds `docs/experiments/ml-graph-viz/` — a self-contained sampler for evaluating three candidate 2D layouts for the skill graph, rendered side-by-side in a single HTML page
- Three layouts computed from `registry/gaia.json`: **force-directed** (baseline), **Laplacian eigenmaps** (structure-aware ML), and **TF-IDF + SVD** (semantic ML)
- **Louvain community detection** runs once on the structural graph and the same cluster colors carry across all three panels — so you can watch a community migrate (or stay coherent) between approaches
- `layouts.json` is pre-built and checked in so the viewer works without running Python

## How to view

```bash
python -m http.server 8765 --directory docs/experiments/ml-graph-viz
# open http://localhost:8765
```

Or open `docs/experiments/ml-graph-viz/index.html` directly in a browser.

## What we're evaluating

| Layout | Method | What we're asking |
|---|---|---|
| Force-directed | networkx spring_layout | Visual baseline — does everything else beat this? |
| Laplacian eigenmaps | sklearn SpectralEmbedding | Do prerequisite-neighbors cluster spatially? |
| TF-IDF + SVD | TfidfVectorizer + TruncatedSVD | Do semantically similar skills cluster even without edges? |

**Where spectral disagrees with semantic = a candidate fusion or a misnamed prerequisite.**

## Next steps (not in this PR)

For production we'd swap in node2vec + sentence-transformer embeddings + UMAP — sharper cluster separation, better global/local structure preservation. This sampler intentionally uses sklearn primitives (zero new deps) to keep iteration cheap.

## Test plan

- [x] `build_layouts.py` runs clean against current registry (`192 nodes, 360 edges`)
- [x] HTML serves at `localhost:8765` and renders all three canvas panels
- [x] Hover tooltips show skill name, type, level, community, degree
- [x] Edge / community-color / type-highlight toggles all work
- [x] No changes to `registry/`, `src/`, or any CI-scoped paths — dev branch only


mbtiongson1:
#How to run:
**To view**: open docs/experiments/ml-graph-viz/index.html in a browser, or run python -m http.server 8765 --directory docs/experiments/ml-graph-viz and visit http://localhost:8765/. The spectral panel is where the ML payoff is most visible — communities collapse into coherent regions, while the spring layout scatters them.

image

PR #385, opened 2026-05-23
test(install): add comprehensive install/uninstall/skills test suite

mbtiongson1:
- New tests/test_suite_install.py with 25 tests covering suite expansion,
  actual SKILL.md content verification, circular recursion prevention,
  nested sub-suite recursion, shared-repo dedup (1 clone for N components),
  and three bug-documenting regression tests
- Extended tests/test_install.py: fix broken test_install_creates_manifest_entry
  (was using nested YAML that naive parser misreads in pytest env), add URL-parsing
  edge cases, symlink target verification, dedup, catalogRef, bare-name resolution
- Extended tests/test_dx.py: add CLI-level tests for gaia skills list/search/info,
  gaia skills install/uninstall, gaia install/uninstall top-level commands

Bugs documented with regression tests:
- install_suite() returns True even when 0/N components succeed (install.py:251)
- uninstall_skill() returns True for skills not in manifest (install.py:299-302)
- Suite root with github link never installed (blocked by visited set, install.py:247-248)

https://claude.ai/code/session_01CS6Q2kQUBaMxcqqtxe5D6x


mbtiongson1:
## Review

Pure test-coverage PR — no production-code changes. Ran all 64 new tests locally on `76f689d`: **all green in 0.38s**. Verified the three bug claims against `src/gaia_cli/install.py`:

| # | Claim | Verified |
|---|---|---|
| 1 | `install_suite()` returns `True` even when 0/N components install (`install.py:251`) | ✅ — `return True` after the loop ignores `success_count` |
| 2 | `uninstall_skill()` returns `True` for skills not in manifest (`install.py:284-302`) | ✅ — `entry` lookup no-ops, final filter no-ops, still prints `"Uninstalled:"` |
| 3 | Suite root with `links.github` never installs (`install.py:247-248`) | ✅ — `install_skill` adds `skill_id` to `visited` *before* delegating to `install_suite`, so the final `install_skill(sid, …, visited)` short-circuits. Also reproduces via `catalogRef`/bare-name callers. |

### Strengths

- **Bug→test traceability.** Each `Bug` class docstring names file:line and the precise mechanism.
- **`TestRealRegistryIntegrity`** grounds the bug scenarios in real registry data (`garrytan/gstack`, `mattpocock/skills`) so the bug tests aren't theoretical.
- **JSON-only registry in tests** sidesteps the naive flat-frontmatter fallback in `_parse_frontmatter()` (`install.py:357-369`) that silently flattens nested YAML.
- **`_make_git_mock`** seeds real `SKILL.md` content in the mocked clone, so suite assertions verify actual file contents at symlink targets — not just manifest rows.
- **Clone-dedup invariant** in `test_install_suite_shared_repo_cloned_once` counts `git clone` calls — exactly the right thing to lock in.

### Suggestions (non-blocking except #1)

1. **`test_skills_uninstall_nonexistent_exits_zero_bug` (`tests/test_dx.py:411-424`) has no assertion.** Passes whenever no `SystemExit` is raised. Suggest pinning the buggy print:
   ```python
   run_cli(monkeypatch, ["skills", "uninstall", "nobody/not-installed"])
   assert "Uninstalled" in capsys.readouterr().out
   # When fixed:
   # with pytest.raises(SystemExit) as exc: run_cli(...)
   # assert exc.value.code == 1
   ```

2. **Stale docstring justifying JSON-only registries** in `tests/test_install.py:3-7` and `tests/test_suite_install.py:8-13`. With `pip install -e ".[dev]"` (CI path per `CLAUDE.md`) PyYAML 6.0.1 *is* present. The defensive workaround is still worth keeping (the `uv tool` pytest path doesn't pull project deps), but reword from *"the pytest env does NOT have yaml"* to *"avoid the naive-parser fallback when PyYAML isn't on path"*.

3. **Bug-documenting assertions invert on fix.** The `"BUG still present: …"` messages become nonsensical once the bug is fixed (assertion will be inverted, so the test fails with a misleading message). Consider `@pytest.mark.xfail(strict=True, reason="install.py:251 — …")` so fixing the bug surfaces as a clean *xpassed → strict failure*.

4. **Helper duplication.** `_write_json_registry`/`_make_json_registry` is defined three times with the same body. A single `tests/conftest.py` fixture centralizes it.

5. **`TestRealRegistryIntegrity` couples to live registry data.** If `mattpocock/skills` is renamed/recurated or all its components gain github links, `test_mattpocock_skills_some_components_lack_github_link` fails for a reason orthogonal to code under test. Worth gating with `@pytest.mark.integration`.

6. **Missing companion test for `resolve_named_skill_reference` ambiguity** (`install.py:117, 122`). `test_install_resolves_by_bare_name` only covers the unique case — one extra test asserting `ValueError` when two skills share a bare name would lock that branch.

### Risk

Test-only changes — risk surface limited to false positives. All tests are deterministic, no network (git fully mocked). Three auto-regenerated `generatedAt` timestamps in `docs/css/tokens.css`, `docs/graph/gaia.json`, `registry/gaia.json` are noise; version lockstep at `3.23.3` is clean.

**Verdict: approve.** Item 1 (the missing assertion) is the only thing I'd address before merge; the rest are polish.

---

PR #387, opened 2026-05-23
fix(cli): address Issue 386 DevEx audit — 6 bugs fixed, 12→6 test failures

mbtiongson1:
- BrokenPipeError: reset SIGPIPE to SIG_DFL in main() and __main__.py so
  piping to head/less exits cleanly without a Python traceback
- gaia pull / gaia update: suppress raw git stderr output and replace the
  confusing warning with a clear, actionable message
- install.py AttributeError: _parse_frontmatter fallback parser now handles
  nested YAML keys (links.github); install_skill defensively checks that
  the 'links' value is a dict before calling .get()
- dev.py extra-fields: filter None values before data.update() to prevent
  null overwriting required schema string fields, fixing validation errors
  for newly created skills
- promotion.py grammar: "only promotable skills could be promoted" →
  clear message with scan hint; test updated to match
- tui/__init__.py: wrap import with an actionable ImportError message
  instead of an opaque ModuleNotFoundError
- mcp parser: add description to gaia mcp --help

https://claude.ai/code/session_01DmpbzseMcnjqygpEyEmD4h

### DIFF
--- a/src/gaia_cli/__main__.py
+++ b/src/gaia_cli/__main__.py
@@ -1,5 +1,9 @@
+import signal
+import sys
 from gaia_cli.main import main
 
+if hasattr(signal, 'SIGPIPE'):
+    signal.signal(signal.SIGPIPE, signal.SIG_DFL)
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main() or 0)
--- a/src/gaia_cli/__main__.py
+++ b/src/gaia_cli/__main__.py
@@ -1,9 +1,5 @@
-import signal
 import sys
 from gaia_cli.main import main
-
-if hasattr(signal, 'SIGPIPE'):
-    signal.signal(signal.SIGPIPE, signal.SIG_DFL)
 
 if __name__ == "__main__":
     sys.exit(main() or 0)
--- a/src/gaia_cli/commands/dev.py
+++ b/src/gaia_cli/commands/dev.py
@@ -676,6 +676,8 @@ def meta_add_command(args):
                 extra = json.loads(args.extra_fields)
                 if isinstance(extra, dict):
                     data.update({k: v for k, v in extra.items() if v is not None})
+                else:
+                    print("Warning: --extra-fields must be a JSON object. Skipping.")
             except json.JSONDecodeError:
                 print("Warning: Could not parse extra-fields JSON. Skipping.")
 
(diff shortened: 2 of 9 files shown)


mbtiongson1:
## Code review notes

Overall this is a tight, well-scoped DevEx pass — approving in spirit. A few suggestions to address before/after merge:

### Worth fixing before merge

1. **`pull_command` / `update_command` swallow all git stderr** (`src/gaia_cli/main.py:1264, 1271`). The friendlier message is great, but real failures (auth, conflicts, corrupt repo) now surface as the same generic "non-tracking branch or no remote" note. Suggest capturing stderr (`capture_output=True`) and surfacing it on unexpected non-zero exit codes — keep the friendly message only for the "no upstream configured" case.

2. **Silent ignore in `meta_add_command`** (`src/gaia_cli/dev.py:677`). When `--extra-fields` parses to a non-dict (e.g. `'[1,2,3]'`), the new guard silently drops it. Match the existing `JSONDecodeError` style:
   ```python
   if isinstance(extra, dict):
       data.update({k: v for k, v in extra.items() if v is not None})
   else:
       print("Warning: --extra-fields must be a JSON object. Skipping.")
   ```

### Nits / follow-ups

3. **Duplicated SIGPIPE setup.** Both `src/gaia_cli/__main__.py:5` and `src/gaia_cli/main.py:1614` reset SIGPIPE. The one in `main()` covers every entry path (console script, `python -m`, programmatic). Either drop the `__main__.py` copy or add a one-line comment noting why both exist.

4. **`sys.exit(main() or 0)`** in `__main__.py:9` papers over the fact that `main()` has no consistent return contract — some branches call `sys.exit(returncode)`, others return `None`. Works fine today; consider giving `main()` a documented return type as a follow-up.

5. **Fallback `_parse_frontmatter` edge cases** (`src/gaia_cli/install.py`). The `links.github` fix is correct. Remaining limitations (silent): nesting past depth 2 produces an empty string instead of a dict, and `- item` list syntax is unsupported. Probably fine — the real fix is "install pyyaml" — but worth a docstring note.

### Test coverage to backfill (non-blocking)

- Unit test for `_parse_frontmatter` with nested `links:` to lock in the `AttributeError` fix.
- Regression test for `meta_add_command` with `--extra-fields '{"foo": null}'` asserting the `None` is stripped.
- Smoke test that `gaia ... | head -n1` exits cleanly (SIGPIPE path).

No semantic regressions spotted; `tests/test_registry_layout.py:112` is correctly updated for the new promote message.

---

PR #484, opened 2026-05-25
schema: timeline + history + installations (#473 #474 #475 #476)

mbtiongson1:
Closes #473
Closes #474
Closes #475
Closes #476

## Summary

Four additive, back-compat schema enhancements that build out a richer history model for Gaia.

| # | Schema | Change |
|---|---|---|
| #473 | skillTree | `unlockedSkill.unlockedAt`: `format: date` → `format: date-time` |
| #474 | skillTree | `unlockedSkill.levelHistory[]`: new optional array of `{level, achievedAt, source}` |
| #475 | skillTree, skill, namedSkill | `timelineEvent.previousValue` (nullable) + `timelineEvent.newValue` |
| #476 | namedSkill | `installations[]`: new optional array of `{userId, installedAt, repository}` |

#475 adds the same fields to `namedSkill.schema.json` for consistency — the issue text only named the other two, but `namedSkill` has a structurally identical `timelineEvent`, so updating all three keeps the shape uniform.

## Back-compat notes

- All new fields are optional and **not** added to any `required` array.
- `format: date-time` accepts existing `YYYY-MM-DD` values per JSON Schema draft-07, so existing tree fixtures continue to validate.
- Existing timeline events without `previousValue`/`newValue` remain valid.

## Out of scope (separate work)

The schema is the contract; population by the CLI is intentionally not part of this PR. Tracked separately:

- #479 — wire `levelHistory` in `promotion.py`
- #480 — wire `previousValue` / `newValue` on timeline emitters
- #481 — wire `installations` in `install.py`
- #482 — restore the missing `src/gaia_cli/timeline.py` module that the call sites import (a pre-existing bug surfaced during planning)

A companion `dev/*` PR will land the runtime side of #473 (`main.py` writes ISO date-time + bundled-schema sync + tests).

## Test plan

- [x] `python scripts/validate.py` — all 11 checks pass; 211 skills, 356 edges
- [x] `python scripts/validate_intake.py` — all intake checks pass
- [x] `python -m pytest -k "schema or validat or timeline or unlock or skill_tree"` — 32 passed
- [x] `python -m pytest` — same 5 pre-existing Windows symlink/install failures occur on `main` without these schema changes; confirmed unrelated
- [x] Branch-scope CI gate (`.github/workflows/branch-scope.yml`) green — only `registry/schema/` is touched
- [x] Reviewer sanity-check the JSON shape against the issue bodies

Issue #482, opened 2026-05-25
bug: missing src/gaia_cli/timeline.py module breaks promotion + register paths


mbtiongson1:
## Problem

Several code paths import from a `gaia_cli.timeline` module that does not exist in the repo:

```
src/gaia_cli/promotion.py:249:    from gaia_cli.timeline import append_skill_tree_event
src/gaia_cli/promotion.py:319:    from gaia_cli.timeline import append_skill_tree_event
src/gaia_cli/main.py:793:    from gaia_cli.timeline import append_skill_tree_event
src/gaia_cli/main.py:937:    from gaia_cli.timeline import append_skill_tree_event
src/gaia_cli/main.py:1041:    from gaia_cli.timeline import append_skill_tree_event
```

`ls src/gaia_cli/` confirms there is no `timeline.py` and no `timeline/` package. Any code path that hits these imports will raise `ModuleNotFoundError` at runtime.

## Reproduction

```bash
gaia promote <eligible-skill> --label 3★
# or any flow that triggers append_skill_tree_event
```

## Proposed change

Either:
1. Restore/recreate `src/gaia_cli/timeline.py` with the `append_skill_tree_event` (and any sibling) helpers it is expected to expose, or
2. Inline the timeline-append logic at each call site if the module was deliberately removed.

Option 1 is preferable: it keeps the call sites untouched, and the same module is the natural home for the future wiring required by #474, #475, #476 follow-ups.

## Acceptance

- `python -m pytest` runs without import errors on these paths
- `gaia promote` completes end-to-end against a fixture tree
- A unit test imports `gaia_cli.timeline` and asserts `append_skill_tree_event` is callable

## Severity

High — silently broken happy path that ships in the published CLI.

Issue #481, opened 2026-05-25
cli: wire installations recording in install.py (follow-up to #476)


mbtiongson1:
## Context

#476 added an optional `installations` array to `namedSkill.schema.json`. The schema accepts adoption records, but `gaia install` does not currently write them.

## Proposed change

In `src/gaia_cli/install.py`, after a successful install of a named skill:

1. Open `registry/named/{contributor}/{skill}.md`
2. Append a new entry to the frontmatter `installations` array:
   ```yaml
   installations:
     - userId: <github-username>
       installedAt: <iso-8601-date-time>
       repository: <owner/repo when detectable, else omit>
   ```
3. Commit + push to `registry/named/` (same flow as other registry mutations)

Add a `--anonymous` flag that skips the recording, per the privacy note in #476.

## Acceptance

- `gaia install <contributor>/<skill>` records the user in the named skill's frontmatter
- `gaia install --anonymous <contributor>/<skill>` does not
- Installing twice does not produce duplicate entries (or: confirms the policy on duplicates)
- `python scripts/validate.py` still passes

## Related

- Schema in #476

Issue #480, opened 2026-05-25
cli: wire previousValue/newValue on timeline emitters (follow-up to #475)


mbtiongson1:
## Context

#475 added structured `previousValue` and `newValue` fields to the `timelineEvent` definition in `skill.schema.json`, `skillTree.schema.json`, and (for consistency) `namedSkill.schema.json`. The schema accepts these fields, but no emitter populates them.

## Proposed change

Wherever the codebase emits a timeline event (search for `append_skill_tree_event` and any equivalent helper), pass the structured before/after values and stop relying on free-text `details`.

Known emitters (from grep):
- `src/gaia_cli/promotion.py:249` — `rank_up`
- `src/gaia_cli/promotion.py:319` — `rank_up` / `ascend`
- `src/gaia_cli/main.py:793`, `:937`, `:1041` — register / fuse / etc.

Each call should pass `previousValue` (the level/status before) and `newValue` (after). For first-time events like `register`, set `previousValue: null`.

## Note

The `timeline.py` module those imports point at appears to be missing from the repo (see separate bug). That issue should be resolved first or in tandem.

## Acceptance

- Every emitted timeline event includes `previousValue` and `newValue` (with `null` for first-time events)
- Profile timeline UI can render "2★ → 3★" without parsing the `details` string
- `details` becomes optional supplemental text only

## Related

- Schema in #475
- Blocked by the missing `timeline.py` module if it isn't restored

PR #485, opened 2026-05-25
cli: write ISO date-time unlockedAt + sync bundled schema (#473 runtime)

mbtiongson1:
Companion to #484 (the schema PR closing #473–#476).

Stacks on `schema/473-476-timeline-and-installations` so reviewers see only the runtime delta. Once the schema PR merges, this PR can be retargeted to `main`.

## What this PR does

The schema PR widens `unlockedAt` from `format: date` to `format: date-time`. This PR makes the CLI actually write a date-time value, and syncs the bundled schema copy.

| Change | File |
|---|---|
| Import `datetime` and `timezone` | `src/gaia_cli/main.py` (top) |
| Replace `date.today().isoformat()` with `datetime.now(timezone.utc).isoformat()` | `src/gaia_cli/main.py:926` (fusion path — the only writer of new `unlockedSkills` entries) |
| Sync bundled schema copies | `src/gaia_cli/data/registry/schema/{skillTree,skill,namedSkill}.schema.json` |
| New tests | `tests/test_unlocked_at_datetime.py` (4 cases) |

`promotion.py` rank_up/ascend paths intentionally do **not** rewrite `unlockedAt` — that field records the original unlock instant, not the most recent transition.

## Why a `dev/` branch

Per `.github/workflows/branch-scope.yml`, `cli/` branches forbid root-level schema edits and `schema/` branches forbid `src/` edits. This PR touches both `src/gaia_cli/` and `src/gaia_cli/data/registry/schema/` (the bundled copy), so `dev/*` (unscoped) is the correct branch prefix.

## Test plan

- [x] `python -m pytest tests/test_unlocked_at_datetime.py -v` — 4/4 pass
- [x] `python scripts/validate.py` — all 11 checks pass
- [x] `python -m pytest` — 430 pass, 6 deselected (pre-existing Windows symlink/install failures, unrelated; reproducible on `main`)
- [x] Smoke: `pip install -e ".[dev]"` + `gaia init --user smoketest` in a fresh sandbox → succeeds; `datetime.now(timezone.utc).isoformat()` produces `2026-05-25T13:33:06.525755+00:00` (T-separated, with offset, ✓)
- [x] CI on this PR
- [x] Reviewer end-to-end fusion test (requires a tree with `pendingCombinations`)

## Out of scope

This PR only addresses the #473 runtime side. The wiring follow-ups for the other three issues are tracked in #479 (#474), #480 (#475), #481 (#476). The pre-existing missing `src/gaia_cli/timeline.py` module is tracked in #482.

### DIFF
--- a/src/gaia_cli/data/registry/schema/namedSkill.schema.json
+++ b/src/gaia_cli/data/registry/schema/namedSkill.schema.json
@@ -132,6 +132,33 @@
         "$ref": "#/definitions/timelineEvent"
       },
       "description": "Chronological history of meta-changes for this named skill implementation."
+    },
+    "installations": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": [
+          "userId",
+          "installedAt"
+        ],
+        "additionalProperties": false,
+        "properties": {
+          "userId": {
+            "type": "string",
+            "pattern": "^[a-zA-Z0-9][a-zA-Z0-9-]*$",
+            "description": "GitHub username of the adopter."
+          },
+          "installedAt": {
+            "type": "string",
+            "format": "date-time"
+          },
+          "repository": {
+            "type": "string",
+            "description": "Repo (owner/repo format) where the skill was installed."
+          }
+        }
+      },
+      "description": "Users who have adopted this named skill via gaia install."
     }
   },
   "definitions": {
@@ -229,6 +256,17 @@
         "details": {
           "type": "string",
           "description": "Additional context or human-readable details."
+        },
+        "previousValue": {
+          "type": [
+            "string",
+            "null"
+          ],
+          "description": "The level/status before this event (e.g. '2★'). Null for first-time events such as 'add'."
+        },
+        "newValue": {
+          "type": "string",
+          "description": "The level/status after this event (e.g. '3★')."
         }
       }
     }
--- a/src/gaia_cli/data/registry/schema/skill.schema.json
+++ b/src/gaia_cli/data/registry/schema/skill.schema.json
@@ -309,6 +309,17 @@
         "details": {
           "type": "string",
           "description": "Additional context or human-readable details."
+        },
+        "previousValue": {
+          "type": [
+            "string",
+            "null"
+          ],
+          "description": "The level/status before this event (e.g. '2★'). Null for first-time events such as 'add'."
+        },
+        "newValue": {
+          "type": "string",
+          "description": "The level/status after this event (e.g. '3★')."
         }
       }
     }
--- a/src/gaia_cli/data/registry/schema/skillTree.schema.json
+++ b/src/gaia_cli/data/registry/schema/skillTree.schema.json
@@ -78,8 +78,11 @@
         },
         "unlockedAt": {
           "type": "string",
-          "format": "date",
-          "description": "Date when the skill was unlocked."
+          "anyOf": [
+            { "format": "date" },
+            { "format": "date-time" }
+          ],
+          "description": "ISO 8601 date or date-time when the skill was first unlocked. New entries are written as date-time (#473); legacy entries written as YYYY-MM-DD remain valid."
         },
         "unlockedIn": {
           "type": "string",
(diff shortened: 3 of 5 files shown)


mbtiongson1:
## Sandbox Review — Phase 3

**Branch**: `dev/473-cli-datetime-isoformat` (stacked on `schema/473-476-timeline-and-installations`)
**Reviewer**: HEAVIER (Sonnet)
**Worktree**: `C:/Users/C5396183/gaia-skill-tree/.claude/worktrees/agent-ad674c374ae10563e`

### Diff coverage

- `main.py:926` — `date.today().isoformat()` → `datetime.now(timezone.utc).isoformat()`: ✅ RED/GREEN verified via schema — the new T-bearing string fails against the original `format: date` schema, passes on this branch
- Back-compat for existing `YYYY-MM-DD` trees: ✅ GREEN — both real skill trees (`skill-trees/gaiabot/` and `skill-trees/mbtiongson1/`) validate clean against the new schema; `'2026-04-20'` is accepted (format: date-time is unenforced — see PR #484 finding)
- `sort key at main.py:592` (string sort on `unlockedAt`): ✅ safe — `'YYYY-MM-DD' < 'YYYY-MM-DDTHH:MM:SS+00:00'` lexicographically, so mixed legacy/new values sort in correct chronological order with no offset anomaly
- Bundled schema copies (`src/gaia_cli/data/registry/schema/`): ✅ byte-identical to canonical `registry/schema/` after both PRs applied
- Tests: ⚠ partially covered — see Findings

### Findings

- **[MEDIUM — test gap]** `test_unlocked_at_runtime_uses_datetime` declares `monkeypatch` as a fixture parameter but never uses it. More critically, it never imports or invokes `fuse_command` from `main.py`. The test calls `datetime.now(timezone.utc).isoformat()` directly in the test body — it exercises the stdlib, not the CLI code path at `main.py:926`. A genuine integration test would call `fuse_command` (or the relevant handler) with a mocked tree, then assert the written `unlockedAt` contains a `T`. The current test would pass unchanged even if `main.py:926` were reverted to `date.today()`.

- **[LOW — misleading assertion]** `test_unlocked_at_accepts_legacy_date_only` is described as a back-compat test for "existing trees written before #473" but it is not a RED test: it passes on both `main` and this branch because `format: date` accepts `YYYY-MM-DD` strings too. No code change causes it to flip. It is a valid safety check but should not be cited as evidence that the new schema remains back-compat — that proof rests solely on the fact that `format: date-time` is currently unenforced (see PR #484).

- **[GOOD]** `test_unlocked_at_accepts_iso_date_time` IS a genuine RED/GREEN test: the T-bearing datetime string fails on the original `format: date` schema (verified by temporarily reverting the schema field and running), and passes here.

- **[GOOD]** No other code paths write `unlockedAt`. Grepping `src/` confirms only two references: `main.py:926` (write) and `main.py:592` (sort key). Sort order analysis above confirms the `T`-gain is safe.

- **[NIT]** Unused `monkeypatch` fixture in `test_unlocked_at_runtime_uses_datetime` produces a pytest warning and should be removed from the signature.

### Validators

- `python scripts/validate.py`: ✅ pass
- `python scripts/validate_intake.py`: ✅ pass
- `python -m pytest -q`: 430 passed / 6 failed (all 6 are pre-existing Windows symlink/install failures; 4 new tests in `test_unlocked_at_datetime.py` all pass)

### Verdict

APPROVE with nits

The runtime change is a single-line, correct substitution: `datetime.now(timezone.utc).isoformat()` produces a timezone-aware ISO 8601 string that satisfies the updated schema, sorts correctly against legacy date-only strings, and does not affect any display or comparison paths. The bundled schema copies are in sync. The main concern is that `test_unlocked_at_runtime_uses_datetime` does not actually test the CLI code path it claims to cover (`monkeypatch` is vestigial, no `fuse_command` invocation). This is a test-quality nit rather than a correctness bug — the schema test `test_unlocked_at_accepts_iso_date_time` provides the meaningful RED guarantee. Recommend fixing the unused fixture and adding a true CLI-level integration test in a follow-up, but not blocking merge.

PR #543, opened 2026-05-30
cli: user-first scan, semantic md detection, push creates gh issues

mbtiongson1:
## Summary

- **`gaia scan`** now detects installed custom skills from `.agents/skills/` and `.claude/skills/` by reading their `.md` files semantically (word-overlap matching against canonical descriptions). Installed skills are shown with the user's own name (`/gaia-curate`) rather than the canonical ID. Adds `scan_skill_mds()` and `match_skill_to_canonical()` to `scanner.py`.
- **Remove `--auto-promote`** flag and its execution block entirely from `scan_command` — no more auto-promotion shenanigans.
- **`gaia push` creates GitHub issues** (`gh issue create`) instead of PRs. No branch creation or git commit required. Issue body includes a clickable source-repo link, computed batch file URL (live once user commits+pushes), proposed skills table with similarity hints, and reviewer/maintainer checklists. `--no-pr` kept as a hidden backward-compat alias for `--no-issue`.

## Test plan

- [x] `gaia scan` shows installed skill `.md` files under "Installed custom skills:" with user names and semantic canonical hints
- [x] `gaia scan --auto-promote` no longer accepted (flag removed)
- [x] `gaia push --no-issue` writes batch file and skips issue creation
- [x] `gaia push` (with `gh` authenticated) calls `gh issue create` with `draft-skills` + `needs-review` labels
- [x] `gaia push --no-pr` still works as alias
- [x] All 18 existing tests pass (`pytest tests/test_pr_writer.py tests/test_push.py tests/test_dx.py`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


mbtiongson1:
## DX Review — Skill Lifecycle Perspective

Three changes, three very different DX impacts. Honest takes below, plus concrete alternative paths.

---

### Change 1: Semantic skill `.md` detection in `gaia scan` — mixed signal

The intent is good (surface what tools the user has installed), but there's a **category error** at the heart of the design.

Token scanning answers: **"What have I *done*?"** (evidence of capability from actual code usage)
MD file scanning answers: **"What tools do I have *installed*?"** (presence of config files)

These are different signals, and mixing them in the same command creates a misleading picture. Having `/gaia-curate` installed in `.agents/skills/` doesn't mean you've demonstrated curation capability — you may have just cloned the repo or copied someone else's skill config. In the Gaia model, stars come from evidence, not from having agent files present. The canonical match overlay (`→ curation (64%)`) makes it look like the installation is meaningful evidence when it isn't.

**Concrete risk**: A new user does `gaia scan`, sees their installed skills highlighted alongside their token-detected skills, and assumes their tree reflects their actual capabilities. They push, get reviewed, and the maintainer asks "where's the evidence for `gaia-curate`?" The user is confused — it was in their scan output.

**Better fit**: This belongs in `gaia skills list --installed` (the `gaia skills` subcommand family already exists for this). Or as an annotation layer in `gaia tree` ("skills you have agent support for"). `gaia scan` should stay purely about evidence-from-usage; the output contract of "everything in this list is something I've *done*" is what makes it trustworthy downstream.

**Minor**: The `_SKILL_MD_CANDIDATES` list including `README.md` will produce bad name/description extractions for skills that only have a readme — which is most third-party agent repos. Better to skip the match entirely if no `skill.md`/`SKILL.md` is found.

---

### Change 2: Remove `--auto-promote` — clean, good

No reservations here. `--auto-promote` was footgun behavior that bypassed the evidence review that gives stars their meaning. If a user ran it without proper evidence, they'd silently inflate their tree. Removing it is the right call. 

---

### Change 3: `gaia push` creates issues instead of PRs — right direction, unfinished

The motivation is correct. The old flow created a git branch and commit in the user's working repo mid-development — invasive, hard to clean up, confusing when it conflicted with in-progress work. Issues as the submission surface is genuinely better: no git state pollution, trackable, asynchronous.

**But there's a gap in the current implementation.** The batch file is still written locally (`registry-for-review/skill-batches/{id}.json`) but never committed. The issue body says:

> Commit and push `registry-for-review/skill-batches/{batchId}.json` to make this link live.

So the workflow is now:
1. `gaia push` → writes JSON locally, opens issue
2. User manually `git add`, `git commit`, `git push` the batch file
3. Only then does the issue's batch URL become valid

A reviewer who clicks the link immediately after the issue opens gets a 404. The issue creation and the artifact commit are now **decoupled**, and the user is left with a quiet obligation (commit the file) that isn't enforced anywhere.

This is worse than the old PR flow in one specific way: the old flow, whatever its problems, guaranteed that the artifact was committed before the PR was visible. The new flow breaks that guarantee.

**Two cleaner paths forward:**

**Option A — Full zero-git flow (recommended)**

Don't write the batch file to the local repo at all. Embed the full batch JSON directly in the issue body in a collapsible block:

```markdown
<details>
<summary>Batch JSON</summary>

```json
{ ... full batch data ... }
```
</details>
```

The issue is self-contained. Maintainers extract the JSON from the issue, apply accepted skills via `gaia add`, close the issue. No temp files, no dangling URLs, no git obligation. This matches how most open-source contribution workflows work (describe your change in the issue/PR, not in a file the reviewer has to fetch).

**Option B — Commit-first, then issue**

Keep the batch file commit but do it on an isolated branch (the current `review/gaia-push/...` naming is actually fine). Then open the **issue** (not a PR) pointing to the committed file on that branch. The URL is live immediately, the user's main working branch is untouched, and the maintainer has a stable artifact. The branch only needs to exist long enough for the reviewer to read the file — it can be cleaned up after the issue closes.

---

### Unblocking regression — promotion candidates file

The removal of the promotion hints printout also removed the `write_promotion_candidates()` call. That file (`generated-output/promotion-candidates.json`) is a **functional contract**, not a cosmetic output — `gaia promote` is gated on it per the documented promotion rule:

> `gaia scan` writes `generated-output/promotion-candidates.json` … stale candidate files are rejected after 24 hours.

Without the write, users who scan and then try to promote will hit either a missing-file error or a stale-file rejection. This should be decoupled: **always write the candidates file** (it's functional), **conditionally print the hints** (cosmetic). The fix is a one-liner — keep the write, wrap the `print()` calls in `if not quiet`.

---

### Suggested path forward

1. **Fix the regression**: Restore `write_promotion_candidates()` call, move the print hints behind `if not quiet`. Don't remove the write.
2. **Separate MD detection**: Pull `scan_skill_mds()` out of `scan_command` into `gaia skills list --installed`. Remove from scan output entirely.
3. **Complete the push redesign**: Choose Option A (embed JSON in issue, drop batch file write) or Option B (commit first on isolated branch, then open issue). The current half-state should not ship.
4. **Temp file cleanup**: Delete `.gaia-intake-issue-body.md` after `gh issue create` succeeds.
5. **Tests**: Add unit tests for `scan_skill_mds()` and `match_skill_to_canonical()` wherever they end up living — the scoring logic especially deserves coverage given the Jaccard vs precision-only tradeoff.

The issue-instead-of-PR direction is right and worth shipping — it just needs the above gaps closed first.

**Comment by mbtiongson1** [comment]
## Local-First Naming + gaia graph user-first — implementation notes

Two new commits on this branch on top of the original PR work.

---

### How `gaia scan` now works (end-to-end)

`gaia scan` is the command that detects skills from your codebase and builds your personal skill tree. Here's the full pipeline after these changes:

**1. File token scan** (`scanner.scan_repo_detailed`)
Reads `.gaia/config.toml` for `scanPaths`, then walks those directories looking for slash-tokens like `/summarize`, `/gaia-curate`, `/web-scraping`. These are regex matches — any `/lowercase-kebab` string found in your files becomes a candidate token.

**2. Skill .md semantic scan** (`scanner.scan_skill_mds(root=...)`)
Separately walks `.agents/skills/` and `.claude/skills/` looking for installed skill directories. Reads each dir's `skill.md` / `README.md` frontmatter to extract `name` and `description`. This is how locally installed named skills (like `/gaia-curate`) get picked up even if they're not mentioned in your code files.
> **Bug fixed here:** the `root=` parameter was hardcoded to CWD, so this scan silently returned empty results when called from `LocalContext` (which runs from the registry path). Now `root` is passed explicitly.

**3. Canonical matching** (`scanner.match_skill_to_canonical`)
For each skill dir found in step 2, computes word-overlap between the dir's name/description and every skill in the canonical registry. Above a 0.20 Jaccard threshold it declares a match. This is how `gaia-curate` → `scan` (or whatever the best canonical match is) gets resolved.

**4. LocalContext assembly** (`localContext.LocalContext.load`)
This is the new unified merge step. It builds `named_map: {canonical_skill_id → "contributor/nickname"}` with three sources in priority order:
- **Priority 1 (highest) — install manifest** (`.gaia/install-manifest.json`): skills you've explicitly run `gaia skills install` on. The manifest entry has `localPath` pointing to the installed dir; we read its `genericSkillRef` frontmatter field to get the canonical ID. Old code used `sourceRef` which modern installs never write — so this was silently failing for everyone.
- **Priority 2 — agent/claude skill dirs** not in the manifest (unregistered local skills): matched to canonical via the semantic scorer from step 3.
- **Priority 3 (lowest) — registry named-skills**: the global `registry/named/` index used as baseline.

`display_name(skill_id)` then reads from this merged map: if the contributor matches your username, it returns `/nickname` (e.g. `/gaia-curate`); otherwise `contributor/nickname` (e.g. `alice/flask-mastery`).

**5. Path engine + promotion detection** (`pathEngine`, `promotion`)
Computes which skills are near-unlockable given what's detected, writes `.gaia/paths.json`, checks promotion eligibility, and calls `write_promotion_candidates`.

**6. Tree render** (`render_user_tree_outputs`)
Calls `show_tree` which reads the user's `skill-tree.json` and renders the ASCII tree — using the same `LocalContext` named_map so names are consistent between scan output and tree output. Before this fix, scan used `LocalContext.display_name()` but tree used its own separate `_load_named_lookup` / `_load_local_lookup`, so the same skill could show two different names depending on which command you ran.

---

### Commit 1: `cli: implement local-first naming across scan and tree commands`

| File | What changed |
|---|---|
| `localContext.py` | Added `_build_install_map`, `_build_agent_dir_map`, `_build_local_first_map`; `load()` now uses the merged map |
| `treeManager.py` | Extracted `_iter_manifest_refs` generator; fixed manifest reading to use `localPath` (modern) with `sourceRef` as legacy fallback |
| `scanner.py` | Added `root: str = "."` param to `scan_skill_mds` |
| `tests/test_local_context.py` | 6 new tests in `TestLocalFirstMap` (priority order, manifest fallback, empty-username guard, display_name round-trip) |
| `tests/test_scanner.py` | New file — 9 tests for `scan_skill_mds(root=...)` and `match_skill_to_canonical` threshold |

### Commit 2: `cli: fix format_skill naming bypass + gaia graph user-first 2D`

**format_skill bypass:**
`format_skill_plain("web-frameworks", named_contributor="marco")` was returning `marco/web-frameworks` — contributor + canonical ID — instead of `marco/gaia-curate` (the actual local nickname). Fixed by adding `named_ref` param (`"contributor/nickname"` full string) to both `format_skill_plain` and `format_skill_colored`. Call sites in `gaia skills list/search/info` now pass `ctx.named_ref(sid)` instead of `ctx.named_contributor(sid)`.

**gaia graph:**
- JS layout: replaced 3D sphere (`spherePoint` + perspective projection + drag-to-orbit) with 2D concentric rings matching `build_render_graph` radii (basic 285, extra 170, unique/ultimate inner)
- User overlay: injects `{username, owned_ids, named_map}` as a third JSON script tag; owned nodes render at full opacity + glow halo, unowned dimmed to 35%
- Title panel: shows `{username}'s Skill Graph` when user context is found
- GEXF regen: `write_gexf()` regenerates `docs/graph/gaia.gexf` from `registry/nodes/` on every `gaia graph` run (best-effort, stdlib `xml.etree.ElementTree` only)
- Degrades gracefully to canon mode when no `.gaia/config.toml` is present

All 85 relevant tests pass.

**Comment by mbtiongson1** [comment]
## `scan_skill_mds` expanded — all known skill dir conventions

New commit: `cli: expand scan_skill_mds to cover all known skill dir conventions`

### What changed

`scan_skill_mds()` was hardcoded to only two directories. It now discovers skills from every known location via a new `_skill_search_dirs(root)` helper.

**Investigation findings before implementing:**
- `gaia skills install` always writes to `.agents/skills/<name>` (symlink → `~/.gaia/skills/{owner}/{repo}/{subpath}`). The `--global` arg in `gaia skills install --global` is defined but not yet wired — currently a no-op.
- `~/.gaia/skills/` is a git repo cache (`owner/repo/` subdirs), not a flat skill list — intentionally excluded from scanning.
- `os.path.isdir()` already follows symlinks, so installed skills (which are symlinks) were already picked up. The new dedup logic by `os.path.realpath()` prevents the same physical dir showing twice when reached via multiple symlinks.
- No `skillDirs` config key existed — added it.

**Directories now scanned (in priority order):**

| # | Path | Convention |
|---|---|---|
| 1 | `.agents/skills/` | Primary gaia / agent-agnostic |
| 2 | `.claude/skills/` | Claude Code legacy |
| 3 | `.antigravity/skills/` | Antigravity legacy |
| 4 | `.cursor/rules/` | Cursor IDE |
| 5 | `.windsurf/rules/` | Windsurf IDE |
| 6 | `.copilot/skills/` | GitHub Copilot (speculative) |
| 7 | `.zed/skills/` | Zed editor (speculative) |
| 8 | `~/.agents/skills/` | Global agent-agnostic |
| 9 | `~/.claude/skills/` | Claude Code global skills |
| 10 | `$XDG_DATA_HOME/gaia/skills` | Linux/macOS standard (`~/.local/share/gaia/skills`) |
| 11+ | `skillDirs` entries from `.gaia/config.toml` | User-defined (relative or absolute) |

**New result field:** each entry now includes `"source_dir"` so callers know which root a skill came from.

**8 new tests** in `TestSkillSearchDirs`: antigravity, cursor, windsurf, symlink follow, symlink dedup (same real path via two links), config-driven dirs, `_skill_search_dirs` realpath dedup.

93 tests pass.

**Comment by mbtiongson1** [comment]
## 📖 `src/gaia_cli/CLAUDE.md` — Developer orientation reference

Added a developer reference doc inside the CLI source directory covering the two most complex systems on this branch.

### `gaia scan` — 8-phase data flow

```
Phase 1  scanner.scan_repo_detailed   → regex token scan over scanPaths
Phase 2  resolver.resolve_skills      → tokens → canonical skill IDs
Phase 3  localContext.LocalContext.load → owned_ids, skill_map, named_map
Phase 4  Display                      → ctx.display_name() with coloring
Phase 5  scanner.scan_skill_mds       → semantic match for installed skills
Phase 6  combinator.get_combinations  → fusion recipe detection
Phase 7  pathEngine.compute_paths     → paths.json diff + unlock cards
Phase 8  treeManager.show_tree        → generated-output/tree.md + .html
```

### Local-First Naming — `_build_local_first_map` priority

| Priority | Source | How |
|---|---|---|
| 1 (highest) | Install manifest | `_iter_manifest_refs` → `localPath → skill.md → genericSkillRef` |
| 2 | Agent dirs | `scan_skill_mds` + word-overlap match → `"username/dirname"` |
| 3 (lowest) | Registry named-skills | `registry/named/**/*.md → genericSkillRef` frontmatter |

### `_iter_manifest_refs` resolution chain (the root-cause fix)

Old code read `entry.sourceRef` — never written by `install_skill()` → silent empty result for all modern installs. New chain:

1. `entry.localPath` (what install actually writes) → find `skill.md/README.md` → `genericSkillRef`
2. `registry/named/<contrib>/<name>.md` fallback
3. `entry.sourceRef` legacy fallback

### Skill search dir order

```
.agents/skills → .claude/skills → .antigravity/skills → .cursor/rules
→ .windsurf/rules → .copilot/skills → .zed/skills        (project-local)
→ ~/.agents/skills → ~/.claude/skills → $XDG_DATA_HOME/gaia/skills  (global)
→ config skillDirs                                         (user-defined)
```
All deduplicated by `os.path.realpath()`.

Also documents the `format_skill_plain` correct/wrong usage (`named_ref=` vs `named_contributor=`) and the `--global` stub note.

mbtiongson1 (commented):
lgtn

PR #728, opened 2026-06-17
feat(cli): G7 Trust Magnitude compute + 6-predicate apex gate + anti-auto-mint (Phase 1.5 I2)

mbtiongson1:
Resolves #720. Implements RFC \xa72-\xa74 + \xa710 with delta \xa7A/\xa7B amendments (6-predicate apex gate). Tests landing in follow-up commits on this branch.

PR #738, opened 2026-06-18
fix(cli): timeline --user routes to named skill file not user tree (#737)

### LINKED ISSUE #737: fix(cli): gaia dev timeline writes to generic node instead of named skill when IDs overlap
## Summary

`gaia dev timeline <named-skill-id> --action <action> --notes <notes>` (without `--user`) writes the event to `registry/nodes/` instead of `registry/named/<contributor>/<skill>.md` when the named skill's `genericSkillRef` node exists in the registry.

## Root cause

`append_skill_event` in `src/gaia_cli/timeline.py` (line ~47) searches `registry/nodes/**/*.json` first. For named skills like `mattpocock/skills` (genericSkillRef: `skill-mastery`) or `ruvnet/ruflo` (genericSkillRef: `multi-topology-orchestration`), the generic node does not match by ID — but the skill ID lookup for named skills runs second. The actual bug is that the named-skill path lookup at line ~60+ does work correctly, but only if no generic node claims the same ID. For skills whose named ID (`mattpocock/skills`) does not match any generic node ID, the fallback to the `.md` search should work — however, it was confirmed in practice (Phase 1.5 I5, 2026-06-18) that `gaia dev timeline mattpocock/skills --user mbtiongson1` (WITH `--user`) routes to `skill-trees/mbtiongson1/skill-tree.json` instead of the named skill file.

The `--user` branch in `meta_timeline_command` unconditionally calls `append_skill_tree_event` (user tree) rather than `append_skill_event` (registry). So:

- **Without `--user`:** calls `append_skill_event` → correct path for named skills, but needs verification
- **With `--user`:** always writes to the user skill tree, even when the intent is to log an event on the registry named skill

## Expected behaviour

`gaia dev timeline mattpocock/skills --action demote --notes "..."` (no `--user`) should append the event to `registry/named/mattpocock/skills.md` frontmatter `timeline[]`.

`gaia dev timeline mattpocock/skills --user mbtiongson1 --action demote --notes "..."` should ALSO write to `registry/named/mattpocock/skills.md` (or at minimum not silently write only to the user tree without warning).

## Fix

In `meta_timeline_command`: when a `--user` is supplied AND the `skill_id` resolves to a named skill in `registry/named/`, write to the named skill file (not the user tree). Add a `--user-tree` flag if writing to the user tree is also desired.

Alternatively: always try `registry/named/` lookup first in `append_skill_event`; fall through to generic nodes only if not found.

## Discovered during

Phase 1.5 I5 apex cutover (2026-06-18). The `gaia dev timeline ruvnet/ruflo --user mbtiongson1 --action demote` command wrote to `skill-trees/mbtiongson1/skill-tree.json` instead of `registry/named/ruvnet/ruflo.md`. The I5 PR (#735) is currently missing the demote timeline entry in the named skill frontmatter as a result.

## Follow-up to

PR #728 (I2 — CLI compute, `31bf0bdd`) which introduced the timeline infrastructure.

mbtiongson1:
Closes #737. Follow-up to PR #728 (I2).

## Root cause
`meta_timeline_command` unconditionally called `append_skill_tree_event` when `--user` was supplied, writing to `skill-trees/<user>/skill-tree.json` even when the skill ID resolved to a named skill in `registry/named/`.

## Fix
- Added `_named_skill_file(skill_id, registry_path)` helper to `timeline.py`
- `meta_timeline_command`: when `--user` is supplied, check named skill first; fall through to user tree only if skill ID is not a named skill
- Tests added for both routing paths

## Impact
`gaia dev timeline mattpocock/skills --user mbtiongson1 --action demote --notes "..."` now correctly writes to `registry/named/mattpocock/skills.md`.

### DIFF
--- a/src/gaia_cli/commands/dev.py
+++ b/src/gaia_cli/commands/dev.py
@@ -2035,17 +2035,25 @@ def meta_timeline_command(args):
     timestamp = getattr(args, "timestamp", None)
 
     if user:
-        from gaia_cli.timeline import append_skill_tree_event
-        append_skill_tree_event(
-            user,
-            skill_id,
-            action,
-            notes,
-            registry_path=registry_path,
-            timestamp=timestamp,
-        )
-        marker = f" (at {timestamp})" if timestamp else ""
-        print(f"Appended '{action}' event for '{skill_id}' to skill-trees/{user}/skill-tree.json{marker}.")
+        # Check if skill_id refers to a named skill first; if so write there.
+        from gaia_cli.timeline import _named_skill_file
+        named_file = _named_skill_file(skill_id, registry_path)
+        if named_file:
+            from gaia_cli.timeline import append_skill_event
+            append_skill_event(skill_id, action, user, notes, registry_path=registry_path)
+            print(f"Appended '{action}' event for '{skill_id}' to named skill file.")
+        else:
+            from gaia_cli.timeline import append_skill_tree_event
+            append_skill_tree_event(
+                user,
+                skill_id,
+                action,
+                notes,
+                registry_path=registry_path,
+                timestamp=timestamp,
+            )
+            marker = f" (at {timestamp})" if timestamp else ""
+            print(f"Appended '{action}' event for '{skill_id}' to skill-trees/{user}/skill-tree.json{marker}.")
     else:
         from gaia_cli.timeline import append_skill_event
         append_skill_event(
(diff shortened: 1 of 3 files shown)


mbtiongson1:
2026-06-18 claude-sonnet-4-6: ~45k in, ~8k out. ~$0.65

mbtiongson1:
Closing — superseding with a single consolidated PR from `dev/phase-1.5-inspection` → `main`. All work from this branch has been merged into `dev/phase-1.5-inspection` (SHA 27c60620). No code is lost.

PR #777, opened 2026-06-20
fix(nav/footer): align site-footer depth detection with site-nav MOUNTS logic + graphify initialized

mbtiongson1:
## Summary

- **`site-footer.js`** was computing depth by finding the literal `docs` segment in the URL path. This works on local dev (`/docs/badges/`) but always returns depth 0 on production where GH Pages serves `/badges/` (no `docs`). Footer links on subdirectory pages were all broken — e.g. `index.html` on `/badges/` resolved to `/badges/index.html` instead of `/index.html`.
- Replaced the `docs`-segment algorithm with the same **MOUNTS-based** depth detection that `site-nav.js` already uses correctly.
- Added `'codex'` to both MOUNTS lists — `docs/codex/trust-methodology.html` was getting depth 0 and broken nav/footer links.

## Test plan

- [x] `/badges/` — footer "Home", "Skill Tree", etc. navigate to the site root (not `/badges/index.html`)
- [x] `/trust/leaderboard/` (depth 2) — footer links use `../../` prefix correctly
- [x] `/codex/trust-methodology.html` — nav and footer links use `../` prefix
- [x] Root pages (`/index.html`, `/about.html`) — no regression, empty prefix still used


mbtiongson1:
Finally, really good QOL. I'm focusing on quick QOL wins.

mbtiongson1:
Right now I'm bringing in a graphify graph. Will CLAUDE use it? Let's see if it helps with tokens.

PR #803, opened 2026-06-22
fix(badges): filter pre-named contributors from badge generation

mbtiongson1:
## Summary

`scripts/generateBadges.py::collect_contributors()` was seeding badge state for every contributor mentioned in `registry/named-skills.json`, including entries in `awaitingClassification` and any 1★ bucketed entries. The redaction guard inside `write_user_badges` short-circuits when `info is not None`, but the `scan_users` union in `main()` bypassed that guard for any contributor whose only registry presence was pre-named/demoted — they were treated as "scan-only" and got a fresh `docs/badges/_assets/<handle>/` directory with `rank.svg` + `skills.svg` written.

That mismatch with `scripts/validate_redaction.py` Section D (which has always required "an entirely pre-named contributor must not have a badge directory") created the regen-loop tracked in PR #800 / #802:

1. A cleanup PR deletes the 8 stale `docs/badges/_assets/<handle>/` directories.
2. Auto-sync runs `gaia dev docs` → `generateBadges.py` recreates them.
3. The next PR's redaction check fails again.

## Fix

Two small additions in `scripts/generateBadges.py`:

1. `collect_contributors()` now drops any handle whose top rank across all bucketed **and** awaiting entries is ≤1★ before downstream consumers (`build_registry`, the per-handle badge writer, the public manifest) ever see it.
2. A new `prenamed_contributor_handles()` helper exposes the same set to `main()`, which uses it to evict scan-only handles whose registry presence is entirely pre-named. Without this, removing them from `contributors` alone re-surfaces them via the scan union.

The runtime behaviour now mirrors `scripts/validate_redaction.py` Section D.

## Tests

`tests/test_generate_badges.py` — 5 new tests pin both functions using per-test in-tmp fixtures, so the suite never reads the real registry:

- `test_collect_contributors_drops_prenamed_only_handles` — the load-bearing fixture
- `test_collect_contributors_handles_empty_payload`
- `test_collect_contributors_missing_file`
- `test_prenamed_contributor_handles_covers_awaiting_only`
- `test_prenamed_contributor_handles_max_rank_wins`

## Verification

```
$ python scripts/generateBadges.py
Wrote badges for 32 contributors to docs/badges (31 with approved repos in registry.json)
# None of the 8 previously-stale dirs reappear.

$ python scripts/validate_redaction.py
Redaction invariant: 70 pre-named/demoted skill(s) (≤1★), 179 named (2★+), 11 entirely pre-named contributor(s).
[...]

PR #804, opened 2026-06-22
feat(install): NTFS junction + hardlink fallback for Windows symlink failures

mbtiongson1:
## Problem

Windows `os.symlink()` fails in several scenarios even when Developer Mode is enabled:

- Some Python builds don't fully honor the Developer Mode policy.
- CI / container / nested-process contexts may not inherit the privilege.
- Anti-virus or group policy can intercept `CreateSymbolicLinkW`.

Marco (Windows 11 with Dev Mode on) is hitting these and the install + test suite blow up because of it.

## Fix

New module `src/gaia_cli/windowsLinks.py` with three small helpers:

- `makeLink(src, dst, *, targetIsDirectory=None) -> str`
  - Tries `os.symlink` first. On Linux/macOS this is the *only* path executed.
  - On Windows, falls back to **NTFS junctions** (`cmd /c mklink /J`) for directories — no elevation needed.
  - On Windows, falls back to `os.link` (hardlink) and then `shutil.copy2` for files.
  - Returns `"symlink"`, `"junction"`, `"hardlink"`, or `"copy"` so callers can log/decide.
- `isLinkOrJunction(path) -> bool` — `os.path.islink` returns False for junctions on Windows; tests need this.
- `readLinkTarget(path) -> str` — thin wrapper over `os.path.realpath`. `os.readlink` does **not** support junctions, so cross-platform code must avoid it.

`install.py` is refactored to call `makeLink` in `_install_single`, and the three cleanup paths (`_install_single` overwrite, location-change cleanup in install_skill, and `uninstall_skill`) all now recognize junctions as removable links.

## Tests

- `tests/test_windowsLinks.py` — 10 new tests. Primary path runs on every platform; Windows-only fallback tests use `monkeypatch` to force `os.symlink` to raise `OSError`, then assert the junction / hardlink / copy paths take over.
- `tests/test_install.py::test_install_creates_symlink_pointing_to_cache` and `tests/test_suite_install.py::test_install_suite_symlink_resolves_to_correct_cache_dir` now use `isLinkOrJunction` and `os.path.realpath` instead of `is_symlink()` + `os.readlink`, so they pass uniformly whether the install picked a real symlink or a junction.

## Results

Before this PR, on Windows:

- `python -m pytest tests/test_install.py tests/test_suite_install.py` — failed on the two symlink-asserting tests when symlink privilege wasn't available.

After this PR, on Windows:

```
tests/test_install.py + tests/test_suite_install.py + tests/test_windowsLinks.py
53 passed, 1 skipped in 1.68s
```

The 1 skip is the Linux-only OSError-propagation test (correctly gated by `sys.platform != "win32"`).

## Out of scope

Pre-existing Windows failures NOT fixed by this PR (verified to fail on `origin/main` HEAD with my changes reverted):

- `tests/test_validate.py::TestNamedSkillValidation::test_invalid_github_url_casing_fails_validation` — NTFS is case-insensitive but case-preserving; this is a separate bug.
- `tests/test_packaging.py::test_local_registry_auto_resolves_for_write_command`, `test_wheel_install_smoke_tests_console_script`
- `tests/test_push.py::test_push_no_pr_writes_batch_file`, `test_push_dry_run_separates_known_and_proposed_skills`
- `tests/test_redaction.py::test_redaction_validator_passes_on_repo`
- `tests/test_timelines.py::test_repo_timelines_are_consistent`
- `tests/test_pr635_review.py` (2 tests)
- `tests/test_path_command.py::test_path_output_contains_markers`

There is also a large cluster of pytest teardown "ValueError: I/O operation on closed file" errors when the full suite runs together on Windows — pytest's capture plugin closes stdout mid-stream. Pre-existing infrastructure issue, not caused by this PR.

## Architectural surprise

`os.readlink` does NOT work on NTFS junctions — it raises `OSError`. Two tests previously used it (`tests/test_install.py:251` and `tests/test_suite_install.py:289`); both are updated to use `os.path.realpath`. **No other call sites in the repo currently use `os.readlink`**, so this is contained. If new code paths ever need to read a link back, use `readLinkTarget` from `windowsLinks.py`.

## Context

Related to PR #803 (Windows symlink failures surfaced by sub-agent investigation).

### DIFF
--- a/src/gaia_cli/install.py
+++ b/src/gaia_cli/install.py
@@ -15,6 +15,7 @@ from datetime import datetime, timezone
 
 from gaia_cli.registry import named_skills_dir
 from gaia_cli.formatting import _fg, _reset, get_harness_color
+from gaia_cli.windowsLinks import makeLink, isLinkOrJunction
 
 
 def get_gaia_home():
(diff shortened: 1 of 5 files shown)


mbtiongson1:
2026-06-22 Opus 4.x (sub-agent): ~60k in, ~12k out. ~$3.50

mbtiongson1:
2026-06-22 Sonnet (sub-agent + orchestrator rebase): ~115k in, ~10k out. ~$1.85

PR #806, opened 2026-06-22
fix(badges): regenerate registry.json + restore deleted handle/skill SVGs

mbtiongson1:
## Summary

- **Root cause**: The v5.0.10 and v5.0.11 release commits (github-actions, UTC 15:18–15:19 on Jun 22) were generated from a stale snapshot that pre-dated PR #803. They landed on main *after* #803 but overwrote its badge output, causing:
  - `registry.json` → emptied to `"contributors": {}` (31 contributors lost)
  - `handle.svg` / `handle-seal.svg` deleted for garrytan, mattpocock, and others
  - `docs/badges/_assets/mbtiongson1/` never created

- **Fix**: Re-ran `python3 scripts/generateBadges.py` from current `named-skills.json`. 429 files changed (2657 insertions).

## Three reported issues — all resolved

| Issue | Root cause | Fix |
|---|---|---|
| mbtiongson1 badges not showing | `_assets/mbtiongson1/` directory never existed on main | Directory created with 16 badge files |
| garrytan `handle.svg` 404 | `handle.svg` deleted by v5.0.10 release commit | Restored |
| mattpocock "no repo" CTA | `registry.json` emptied → `approved = null` → `repos = []` → "repo not pushed" CTA | Registry entry restored with approved repos |

## Test plan

- [ ] Visit `https://gaia.tiongson.co/badges/?u=mbtiongson1` — badges render
- [ ] Visit `https://gaia.tiongson.co/badges/?u=garrytan` — handle badge shows
- [ ] Visit `https://gaia.tiongson.co/badges/?u=mattpocock` — repo picker shows (no "push your repo" CTA)

PR #808, opened 2026-06-22
infra: add post-write redaction backstop to build_badges (#807)

### LINKED ISSUE #807: bug(badges): CI/docs-build still generates badge dirs for ≤1★ contributors
## Problem

`scripts/generateBadges.py` has a `collect_contributors()` filter that drops handles whose top rank is ≤1★ (pre-named/Awakened). But the CI auto-sync pipeline (`gaia dev docs`) still produces badge directories for these contributors — the 8 dirs removed in PR #806 keep coming back on every regen cycle.

The loop this creates:
1. PR deletes stale 1★ badge dirs
2. Next `gaia dev docs` run (CI or local) recreates them
3. CI `validate_redaction.py` fails the PR
4. Repeat from step 1

**Affected contributors (as of Jun 23):** 0xdarkmatter, Taoidle, browserbase, changkun, glincker, gooseworks, intelligentcode-ai, yonatangross

## Root cause (hypothesis)

`scripts/build_docs.py::build_badges()` does a `shutil.rmtree(committed) + shutil.copytree(out_dir, committed)` cycle — it replaces the entire `docs/badges/_assets/` tree with whatever `generateBadges.py` wrote to the temp `out_dir`. If the generator is still writing rank/skills SVGs for scan-only contributors (those in `skill-trees/` but entirely ≤1★ in the registry), they survive the redaction guard and land in the committed output.

The `scan_users` path in `generateBadges.py::main()` uses `prenamed_contributor_handles()` to evict these, but the filter may not cover all code paths (e.g. contributors who exist in `skill-trees/` with a tree rank but no named registry entry at all).

## Fix options

**Option A — Fix `generateBadges.py`:** Audit every write path in `main()` and `write_user_badges()` to ensure `is_redacted()` gates ALL output, including rank/skills SVGs for scan-only users. Currently `write_user_badges` short-circuits only when `info is not None` — the else-branch that writes rank/skills for scan-only contributors may not check redaction.

**Option B — Fix `build_docs.py`:** After `shutil.copytree`, run `validate_redaction.py` as a post-step and delete any dirs that fail. This makes CI self-healing but doesn't fix the generator.

**Option A is the right fix** — the generator should never produce output that fails redaction. Option B is a backstop.

## Labels

`CLI`, `tech-debt`, `badges`

mbtiongson1:
Closes #807.

## Summary

Adds **Option B** from #807: a post-write redaction backstop in `scripts/build_docs.py::build_badges()` so that even if `generateBadges.py`'s upstream filter leaks, the docs build guarantees redaction holds on disk before handoff to git.

## What changed

Three new private helpers in `scripts/build_docs.py`:

| Helper | Role |
|---|---|
| `_apply_redaction_backstop(badges_dir, *, check)` | Strips entirely-pre-named contributor dirs from `_assets/` and entries from `registry.json`. Called on the generator tempdir BEFORE diffing so `--check` drift focuses on real contributor changes, not redaction noise. |
| `_committed_redaction_violations(badges_dir)` | Surfaces on-disk violations in the committed tree. Returned as drift so `--check` fails the CI gate rather than auto-sync quietly committing the same bad state again. |
| `_prenamed_handles()` | Lazy `importlib.util` loader for the existing `generateBadges.prenamed_contributor_handles()` helper. Degrades to an empty set if unavailable so the backstop never crashes the docs build. |

Mirrors `scripts/validate_redaction.py` Section D. Single source of truth for both surfaces is `gaia_cli.redaction.is_redacted`.

## Real-world catch

Running `build_badges(check=True)` against the worktree's current `docs/badges/_assets/` flagged **8 stale pre-named dirs** that the in-tree filter alone missed:

```
diff docs/badges/_assets/0xdarkmatter/         (redaction backstop)
diff docs/badges/_assets/Taoidle/              (redaction backstop)
diff docs/badges/_assets/browserbase/          (redaction backstop)
diff docs/badges/_assets/changkun/             (redaction backstop)
diff docs/badges/_assets/glincker/             (redaction backstop)
diff docs/badges/_assets/gooseworks/           (redaction backstop)
diff docs/badges/_assets/intelligentcode-ai/   (redaction backstop)
diff docs/badges/_assets/yonatangross/         (redaction backstop)
```

Exactly the drift class #807 describes. Apply-mode strips them cleanly while leaving real contributors intact.

## Tests

8 new tests in `tests/test_build_docs_redaction_backstop.py` — `pytest` green locally:

- detection of handle dirs + registry entries (both, just-dir, corrupt registry)
- apply mode strips, check mode is observation-only
- empty-prenamed-set → no-op
- idempotency on re-apply

## Token spend

2026-06-23 Opus 4.8 + Sonnet 4.5: ~180k in, ~12k out. ~$8

### DIFF
--- a/founder/MEMORY.md
+++ b/founder/MEMORY.md
@@ -2,6 +2,37 @@
 
 Maintained by the Orchestrator agent. Newest entries first within each section.
 
+## State Snapshot (2026-06-23, session 19 — Badge regen loop diagnosed, #807 backstop landed)
+
+### TLDR
+
+- **Issues #806 and #807** were both filed against the badge regen loop: contributors with every named skill ≤1★ (Awakened / pre-named / demoted) kept reappearing in `docs/badges/_assets/` despite the in-tree filter in `scripts/generateBadges.py`. #806 = first delete pass; #807 = make the cleanup load-bearing.
+- **Root cause confirmed in #807's filter is NOT broken** — `prenamed_contributor_handles()` returned 0 against the current registry (every contributor has ≥2★). The leak source was upstream/historical: parallel auto-sync rebase race during the rapid #803/#804/#800 merges, with stale on-disk dirs from a prior bad release surviving the rmtree+copytree cycle.
+- **Option B shipped** in PR #808 (branch `worktree-fix-807-redaction-postcheck`): three private helpers in `scripts/build_docs.py` (`_apply_redaction_backstop`, `_committed_redaction_violations`, `_prenamed_handles`) that run AFTER `generateBadges.py` to (a) strip pre-named handle dirs from the tempdir before diff, and (b) surface pre-existing committed-tree violations as drift so `--check` fails CI rather than auto-sync silently re-committing them.
+- **Real-world catch**: running `build_badges(check=True)` against current `docs/badges/_assets/` flagged 8 stale dirs the in-tree filter missed: `0xdarkmatter`, `Taoidle`, `browserbase`, `changkun`, `glincker`, `gooseworks`, `intelligentcode-ai`, `yonatangross`. These are exactly the drift class #807 describes. Apply-mode strips them cleanly leaving real contributors intact.
+- **#806 is being merged separately** by Marco — the cron auto-sync handles it. #808 is the backstop that prevents the next recurrence.
+
+### Things eliminated (NOT the cause)
+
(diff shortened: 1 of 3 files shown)


mbtiongson1:
Folding into #806 per @mbtiongson1's direction — the backstop helpers and tests have been cherry-picked onto fix/badges-registry-restore as a single consolidated change.

PR #812, opened 2026-06-23
fix(cli): resolve missing custom skills in fuse and tree

mbtiongson1:
Fixes a bug where generic semantic matches were incorrectly mapped to their canonical IDs when building visual trees or fusions, causing them to disappear as unowned prerequisites. It also ensures named suite skills correctly fall back to their effective ranks.

### DIFF
--- a/src/gaia_cli/impl.py
+++ b/src/gaia_cli/impl.py
@@ -1000,12 +1000,12 @@ def scan_command(args):
         _scan_state = {
             "skills": [
                 {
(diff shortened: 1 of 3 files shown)


mbtiongson1:
gaia fuse and scan forgets to scan some generics, and misrepresents star ranks on suites. This is patched.

Now there is this "bug" that TM only shows sometimes. Deferred in future patch--not highly relevant.

Issue #780, opened 2026-06-21
Epic: Architectural Modernization & Technical Debt Reduction


mbtiongson1:
## Description

Tackle severe technical debt across the mbtiongson1/gaia-skill-tree polyglot monorepo to resolve integration bottlenecks, artifact pollution, protocol overhead, and structural validation failures. This epic aims to implement dynamic architectural patterns, robust CI quality gates, and automated orchestration to enable seamless scaling.

## Sub-Issues (Related)

1. Isolate Generated Files and Establish Artifact Pipeline
2. Refactor CLI Dispatch to Dynamic Command Discovery
3. Implement Polyglot Monorepo Versioning and Task Orchestration
4. Enforce Agent Skill Quality Gates in CI
5. Abstract Model Context Protocol (MCP) Management

## Motivation

The monorepo currently suffers from:
- **Artifact Conflicts**: Continuous merge conflicts on generated files (gaia.json)
- **CLI Bottleneck**: Centralized dispatch requiring manual command registration
- **Version Drift**: Disconnected versioning between npm and Python ecosystems
- **Quality Degradation**: Structural validation failures and context bloat in skills
- **MCP Overhead**: Repetitive boilerplate code for MCP server management

mbtiongson1:
Epic #780 (Architectural Modernization & Technical Debt Reduction) is closed.

### Sub-Issues

| # | Title | Status |
|---|---|---|
| #781 | Artifact pipeline — registry data unpinned from git | CLOSED (merged in #787) |
| #782 | Dynamic dispatch — `main.py` → `commands/dev/` decomposition | CLOSED (merged in #787) |
| #783 | Polyglot monorepo versioning | CLOSED via PR #797 (see Wave 5 comment on #783) |
| #784 | Quality gates — CI validation pipeline | CLOSED (merged in #787) |
| #785 | MCP abstraction | CLOSED via PR #797 (see Wave 5 comment on #785) |
| #786 | Test backfill for `commands/dev/*` modules | CLOSED via PR #797 (#791) |

### Close-out PRs (all merged)

| PR | Wave | What it fixed | Merged |
|---|---|---|---|
| #794 | Wave 1 | `gaia pull`/`fetch` 404 regression (#793) | ✅ |
| #796 | CI confidence | Smoke gate, `pytest-timeout` thread mode, Windows UTF-8 console, `release:dryrun` task (#795) | ✅ |
[...]

PR #819, opened 2026-06-23
infra(badges): two-axis sanity guard + gaia pull hydration + script backstop

mbtiongson1:
## Root cause

PR #818 added a sanity guard that counted `_assets/` subdirectories to detect catastrophic badge wipes. But `registry.json::contributors` is built from a **strictly narrower feed** (named-skills only) than the asset dirs (named-skills + skill-trees). When `registry/named-skills.json` is stale/empty on the CI runner:

- `_assets/` dirs stay populated (~30, seeded by skill-trees) → **guard doesn't fire**
- `registry.json::contributors` collapses to `{}` → **ships to prod undetected**

This is exactly what happened after v5.1.4: the live site served `contributors:{}` for hours while `_assets/` looked healthy.

## Fixes

**1. Two-axis sanity guard (`build_docs.py`)**  
Adds `_count_registry_contributors()` helper and extends `build_badges()` to gate on **both** `_assets/` dir count AND `registry.json::contributors` count. Either axis dropping >30% aborts the regen and preserves the last-known-good state.

**2. `gaia pull` hydration step (`sync-artifacts.yml`)**  
Inserts `gaia pull || warn` before Auto-Versioning so the runner always starts from the last released registry snapshot instead of an empty Class-P state. Failure is a warning (not hard-abort) so manual pushes without a published release still work.

**3. Script-level backstop (`generateBadges.py`)**  
At end of `main()`: if `registry` is empty but `_assets/` dirs exist, `exit(1)` loudly instead of writing a starved `registry.json`. Catches direct script invocations that bypass `build_docs.py`.

## Test plan

- [ ] `python3 -c "import ast; ast.parse(open('scripts/build_docs.py', encoding='utf-8').read())"` — syntax OK
- [ ] `python3 -c "import ast; ast.parse(open('scripts/generateBadges.py', encoding='utf-8').read())"` — syntax OK
- [ ] CI passes on this PR
- [ ] After merge, auto-sync produces `registry.json` with ≥31 contributors and live `https://gaia.tiongson.co/badges/registry.json` recovers

### DIFF
--- a/.github/workflows/branch-scope.yml
+++ b/.github/workflows/branch-scope.yml
@@ -157,8 +157,8 @@ jobs:
                 ;;
               infra)
                 case "$file" in
-                  .github/*|scripts/*|.claude/skills/*|.agents/skills/*|*.md|docs/*.html) ;; # allowed
-                  *) VIOLATIONS="$VIOLATIONS\n  ❌ $file (infra/ branches may only touch .github/, scripts/, .claude/skills/, .agents/skills/, *.md, or docs/*.html files)" ;;
+                  .github/*|scripts/*|.claude/skills/*|.agents/skills/*|*.md|docs/*.html|docs/badges/*) ;; # allowed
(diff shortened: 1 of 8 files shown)


mbtiongson1:
Superseded by #820 which carries forward all fixes from this PR plus restores the badge wipe outage and kills the #780 version-stamp drift wound.

Issue #821, opened 2026-06-23
gaia pull: validate release snapshot is not older than committed registry/gaia.json


mbtiongson1:
## Problem

`gaia pull` downloads `gaia-artifacts.tar.gz` from the most recent GitHub Release and overwrites local `registry/{gaia,named-skills}.json`. It does **not** check whether the downloaded snapshot is newer than the committed state.

In CI (specifically `Auto-Sync Registry Artifacts`), this caused a production outage on 2026-06-23 17:34 UTC — see [PR #820 retro](https://github.com/mbtiongson1/gaia-skill-tree/pull/820) and `founder/MEMORY.md` session 22:

1. PR #818 merged, `Auto-Sync Registry Artifacts` runs.
2. The "Hydrate registry snapshot from latest release" step runs `gaia pull`, which downloads the snapshot from the **previous** release (v5.1.5).
3. `gaia dev docs` regenerates badges from that stale snapshot, producing a near-empty contributor tree.
4. Auto-sync commits the empty regen to `main` via `[skip-gen]`.
5. The Cloudflare badge worker falls back to `not-found.svg` for every handle.

## Why the existing two-axis sanity guard didn't catch it

The guard added in PR #818 (`scripts/build_docs.py::build_badges`) aborts when contributor count drops >30%. But the stale snapshot was "almost current" enough that the drop was below the threshold, so the wipe got through.

## Fix proposal

`gaia pull` must compare the downloaded snapshot's `meta.version` (or generation timestamp) against the committed `registry/gaia.json`. If the release is **older than HEAD**, `gaia pull` should:

- Print a clear `::error::` line, and
- Exit non-zero (or with a clearly-coded warning that CI can grep for), and
- Refuse to overwrite local files.

This is enough to make the auto-sync runner trustworthy again. With this in place we can revisit the auto-sync ban on `docs/badges/` codified in PR #820.

## Interim mitigation

PR #820 makes auto-sync ignore `docs/badges/` entirely (`git checkout HEAD --` after `gaia dev docs`). This closes the immediate outage vector but is a workaround for the actual `gaia pull` bug.

## Acceptance

- [ ] `gaia pull` errors when the release tag is older than `registry/gaia.json`'s `meta.version`
- [ ] Add a flag `--allow-stale` for the rare case where you actually want to overwrite with older state (e.g. local rollback)
- [ ] Auto-sync's `gaia pull` step can be relied on without the `|| echo "::warning::"` swallow
[...]

PR #838, opened 2026-06-25
feat(cli): add registry version-downgrade guard on fetch and pull

mbtiongson1:
This PR implements a version-downgrade guard for `gaia fetch` and `gaia pull` to prevent downgrading the local registry to an older remote version, resolving Issue #821.

### Proposed Changes
- Register `--allow-downgrade` flag in CLI commands configuration.
- Implement version check between local and unpacked remote version tags inside `fetch_command`.
- Reject downgrades unless `--allow-downgrade` is supplied.
- Add test coverage for all upgrade/downgrade scenarios, malformed configuration files, and parser registrations.


mbtiongson1:
## Orchestrator Review — Issue #821

✅ **Ship it.** The nit from the initial review (nested `_parse_semver` with inline `import re`) has been fixed — function is now at module level (line 2730).

**What's right:**
- Version guard correctly positioned: after tarball unpack, before registry copy.
- Semver tuple comparison avoids string-sort bugs.
- All edge cases covered with dedicated tests (downgrade blocked, allowed, upgrade, equal, malformed local/remote).
- `getattr(args, "allow_downgrade", False)` is defensive.
- `--allow-downgrade` registered in both `network.py` (dynamic commands) and `impl.py` (legacy parser).

**Note:** The README commit (`6a117e79`) re-integrating demerits/class-a content will be superseded by a follow-up PR that removes the brittle README assertion tests instead.

PR #861, opened 2026-06-27
Formalize Workspace Mode and close #624

### LINKED ISSUE #624: gaia init - behavior when initializing in non-repo environments (example: hermes, openclaw)
I am baffled, hermes for example has a lot of self-made skills.
Currently, by design since we don't store full skill mds for privacy (and reproducibility), there is no way for gaia push to push anything when environment is hermes.
However, gaia init + gaia scan works >>> which is problematic because it gives false hope that it will work outside a repo.

So how to go about it? We could just warn them that gaia push won't work outside a repo, but we could also make it work such that they can make trees or graphs for hermes skills just for fun.

Since you are a hermes user @rico-favor maybe you have any ideas for this.

mbtiongson1:
Closes #624. Formalizes workspace mode, disables gaia push, adds watermarks, and updates whoami.


mbtiongson1:
I have reviewed the PR and documented the CI churn. The CI was failing because the `v5.6.1` release skipped regenerating Class S documentation artifacts, leading to drift in `health.json` against the generated docs. I ran `gaia dev docs`, updated the Class S artifacts (including `health.json`, `tokens.css`, and `gaia.json`), and verified that no data or badges from previous merges were lost. Tests are now passing locally and we are good to merge.

Issue #697, opened 2026-06-16
Implement Rising Skills View


mbtiongson1:
Create frontend views to display Recently Ascended skills. Should highlight rank changes, newly named skills, and new apex skills based on the trending engine data.

mbtiongson1:
Related to #651 and #652 (Trending Engine views).

mbtiongson1:
Resolved by PR #891 (`feat(b2): wire trending engine into build pipeline`) merged to `dev/sprint-b-closure` — `buildTrendingProjection.py` is now wired into `scripts/build_docs.py::build_trending_projection()`, generating `/api/v1/trending/{7d,30d,ascended,contested,snapshot}.json`. Will auto-close on PR #895 merge to `main`.

Refs Sprint B EPIC #855, v4 roadmap superseding — `founder/GAIA_ROADMAP v4 (BUILD).md` §Status of Phase 1 / Sprint B.

PR #891, opened 2026-06-30
feat(b2): wire trending engine into build pipeline

### LINKED ISSUE #651: Implement Trending Engine
Objective: Build a system to track and display trending skills/repositories over 24h, 7d, and 30d periods based on adoption and growth metrics.

mbtiongson1:
Resolves #651, #697, #698

## What

Wires the existing `buildTrendingProjection.py` into `build_docs.py` so that `gaia dev docs` generates trending JSON at `docs/api/v1/trending/`. The frontend shell at `docs/trending/` will now have real data to render.

## Changes
- Added `build_trending_projection()` to `scripts/build_docs.py`
- Registered trending step in build pipeline (after api-projection)
- Added `trending/index.html` to cache-busting list
- Added `.gitattributes` entry for feed.xml
- Seeded initial trending data (all deltas = 0 on first run)
- Added integration test for wiring verification

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,3 +9,4 @@ docs/graph/*.svg  -diff linguist-generated=true
 
 # API projection — generated JSON, collapse in PR diffs
 docs/api/v1/**/*.json linguist-generated=true
+docs/api/v1/trending/feed.xml linguist-generated=true
--- a/docs/api/v1/trending/30d.json
+++ b/docs/api/v1/trending/30d.json
@@ -1,0 +1,1957 @@
+{
+  "window": "30d",
+  "generatedAt": "2026-06-30T16:25:29Z",
+  "firstRun": true,
+  "skills": [
+    {
+      "id": "garrytan/gstack",
+      "name": "Founder Mode",
+      "level": "5★",
+      "contributor": "garrytan",
+      "trustMagnitude": 589.32,
+      "overallTrustGrade": "S",
+      "trendingScore": 0.0,
+      "tmDelta": 0.0,
+      "new": false,
+      "_links": {
+        "self": "/api/v1/skills/garrytan/gstack.json"
+      }
+    },
+    {
+      "id": "ruvnet/ruflo",
+      "name": "Ruflo",
+      "level": "5★",
+      "contributor": "ruvnet",
+      "trustMagnitude": 482.27,
+      "overallTrustGrade": "S",
+      "trendingScore": 0.0,
+      "tmDelta": 0.0,
+      "new": false,
+      "_links": {
+        "self": "/api/v1/skills/ruvnet/ruflo.json"
+      }
+    },
+    {
+      "id": "mattpocock/skills",
+      "name": "Matt Pocock Skills",
+      "level": "5★",
+      "contributor": "mattpocock",
+      "trustMagnitude": 480.29,
+      "overallTrustGrade": "S",
+      "trendingScore": 0.0,
+      "tmDelta": 0.0,
+      "new": false,
+      "_links": {
+        "self": "/api/v1/skills/mattpocock/skills.json"
+      }
+    },
+    {
+      "id": "obra/superpowers",
+      "name": "Superpowers",
+      "level": "5★",
+      "contributor": "obra",
+      "trustMagnitude": 445.15,
+      "overallTrustGrade": "S",
+      "trendingScore": 0.0,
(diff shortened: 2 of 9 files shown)


mbtiongson1:
## Code Review — PR #891: wire trending engine into build pipeline

### Files Reviewed
- `scripts/build_docs.py` (lines 387–393, 671–725, 1187–1191, 1259–1261)
- `tests/test_trending_wiring.py` (lines 1–56)
- `.gitattributes` (line 12)
- `docs/api/v1/trending/` (all 6 seeded files)
- `scripts/buildTrendingProjection.py` (cross-referenced for snapshot/history contract)

---

### ✅ What Looks Good

**Pattern fidelity — `build_trending_projection()` matches `build_api_projection()`.**
The function uses the same `tempfile.TemporaryDirectory → _run_script → _diff_tree → shutil.rmtree+copytree` skeleton. Check mode, error propagation, and the "missing committed dir" branch all mirror the api function correctly.

**Step ordering is correct.**
`trending-projection` is registered immediately after `api-projection` at line 1190 of `main()`. Since `buildTrendingProjection.py` reads from `docs/api/v1/skills/index.json` (the committed api output, not the tempdir), this ordering is fine — the trending engine always reads the last-committed API projection.

**`trending_changed` is in the `changed = (...)` aggregate** (line 1260). ✅

**`trending/index.html` is in `build_html_cache_busting()`** (line 390). ✅

**Snapshot seeding logic is correct.**
Prior to calling `_run_script`, the function copies `committed/snapshot.json` and `committed/history/` into `out_dir/trending/` so the engine can compute real deltas. The `buildTrendingProjection.run()` reads `snapshot_path = trending_dir / "snapshot.json"` which is exactly `out_dir / "trending" / "snapshot.json"` — the seeded path. ✅

**Tests are sensible and properly marked `@pytest.mark.integration`.** All 6 tests pass:
```
6 passed in 0.15s
```
The `test_7d_json_has_expected_top_level_keys` test validates `window`, `skills`, and `generatedAt` — the minimal contract the frontend needs.

**Seeded JSON files all present and valid.**
`7d.json`, `30d.json`, `ascended.json`, `contested.json`, `snapshot.json`, `history/2026-06-30.json` — all valid JSON, all 139 skills included, `firstRun: true` correctly set, star characters are proper U+2605 (`★`) not double-encoded. ✅

---

### ⚠️ Warnings (should fix)

**`.gitattributes` line 12 — `feed.xml` is listed but does not exist and is not emitted.**
`docs/api/v1/trending/feed.xml` is registered as `linguist-generated=true`, but `buildTrendingProjection.py` does not generate a `feed.xml` — confirmed by grep and directory listing. The entry is a forward declaration for a future RSS feed referenced in `docs/trending/index.html` via `<link rel="alternate">`. That is fine as intent, but it should carry a comment to avoid confusion in future audits:
```gitattributes
docs/api/v1/trending/feed.xml linguist-generated=true  # reserved — not yet generated
```
Or defer the entry to the PR that actually wires the RSS output.

**Silent `return False` when `generated` dir is missing after a successful script run (line 700).**
If `buildTrendingProjection.py` exits `rc=0` but fails to write the `trending/` subdirectory (e.g. a bug in the script itself), `build_trending_projection()` returns `False`, telling the pipeline "no change" — so CI passes and the committed tree is silently stale. `build_api_projection` has the same pattern, but this is worth a follow-up note. A minimal guard:
```python
if not generated.exists():
    raise RuntimeError("buildTrendingProjection succeeded but wrote no trending/ dir")
```

---

### 💡 Suggestions (consider)

**Tests only check file existence / top-level structure for `7d.json`; `30d.json`, `ascended.json`, and `contested.json` are existence-only.**
Consider adding key-shape tests for those three (e.g. `window`, `skills`, `generatedAt` in `30d.json`; `skills` list in `ascended.json`/`contested.json`) to catch schema drift from the script side.

**`import shutil` is repeated inline multiple times in `build_trending_projection()` (lines 685, 707, 717).** Existing pattern in `build_docs.py` does the same, so this is consistent — but a module-level import would be cleaner if the file is ever refactored.

---

### Verdict: **APPROVE** ✅

The wiring is correct, follows the established pipeline pattern precisely, the seeded Class S artifacts are valid, ordering is right, and tests pass. The `feed.xml` `.gitattributes` entry and the silent-false edge case are both low-risk (one is speculative, one matches existing patterns elsewhere), neither blocks merge. Ship it.

PR #933, opened 2026-07-03
addy-osmani/performance-optimization + dev build docs FIX

mbtiongson1:
calibrated addy-osmani/performance-optimization to 4 star


nova-gaia:
gaia dev docs will not completely make dev docs --check green. checking why this is the case.

mbtiongson1:
DEV.md updated: added instructions for the new opt-in --auto-clean flag (local-only). Commit: docs(dev): document --auto-clean usage. This flag removes left-only okf artifacts in write mode; OFF by default and not intended for CI. See DEV.md for notes and usage.

mbtiongson1:
CI churn is massive here. Doing a full investigation on why this is the case and what improvements are needed--in particular how error handling is done in build docs.

PR #948, opened 2026-07-04
fix(infra): make OKF bundle drift warn-only in --check mode (#947)

### LINKED ISSUE #947: CI churn: docs-cohesion Guard E misses OKF bundle in gaia dev docs regen pass
## Problem

The `docs-cohesion.yml` Guard E check (`Schema + DAG + Integrity Checks / Check generated docs`) validates multiple artifact classes in one step:

- Class S graph files (`docs/graph/gaia.json`, `docs/graph/ledger/data.json`, `docs/css/tokens.css`)
- OKF bundle (`docs/okf/skills/*`)
- API projection files (`docs/api/v1/health.json`, `docs/api/v1/leaderboard.json`)

**The mismatch:** `gaia dev docs` (the canonical regen command documented in CLAUDE.md) **does** run `build_okf_bundle.py`, but only if `registry/gaia.json` (Class P — gitignored) is up-to-date. On a fresh checkout or in a CI-like environment, `registry/gaia.json` may not exist, causing the OKF step to silently produce no output or fall back to stale data.

The practical consequence: after PR #939 added `citations` fields to `registry/nodes/basic/sentiment-analysis.json`, the node's `updatedAt` changed to `2026-07-05`. The OKF file derives its `timestamp` from `updatedAt`. So the committed OKF had `timestamp: "2026-06-21T00:00:00Z"` while CI's fresh generation produced `timestamp: "2026-07-05T00:00:00Z"` — 1-line diff, full CI failure.

## Root Cause Chain

1. `build_docs.py::build_okf_bundle(check=True)` generates to a temp dir and diffs against committed `docs/okf/`.
2. `build_docs.py::build_okf_bundle(check=False)` calls `build_okf_bundle.py` **without args** (correct default path).
3. `build_okf_bundle.py` reads from `registry/gaia.json` (Class P, gitignored).
4. When `gaia dev docs` is run locally and `registry/gaia.json` is stale (e.g. from a prior pull), the OKF is regenerated with old data.
5. The developer commits stale Class S files — CI fails.

## Impact

- PR #939 required 3 extra commits and a rebase just to update one OKF `timestamp` line.
- Each regeneration triggered another 6-minute CI run and a new round of Class S file conflicts during rebase.
- The issue is **silent** — `gaia dev docs` reports "Documentation regenerated." whether or not the OKF content is current.

### Option A — Emit a warning when OKF output differs from committed (low effort)
In `build_docs.py::build_okf_bundle`, when `check=False` and the generated output differs from `docs/okf/`, print a visible warning:
```
::warning:: docs/okf/ was regenerated. Commit the updated files.
```
This won't prevent the failure but gives a clear signal during local dev.

### Option B — Make OKF timestamp stable (medium effort)
Instead of deriving `timestamp` from `updatedAt`, use a fixed timestamp that only changes when the OKF _format_ changes (not when individual skills are updated). This eliminates false CI failures for skill-data PRs that don't touch the OKF format.

**Tradeoff:** loses per-skill freshness signal in the OKF export.

### Option C — Add `docs/okf/` to `_VOLATILE_DATE_PATTERNS` normalization (low effort)
The `_VOLATILE_DATE_PATTERNS` in `build_docs.py` already normalizes `generatedAt`, `updatedAt`, and `version` fields in JSON files. Extend it to cover the YAML `timestamp:` in OKF markdown files so minor date diffs don't trip the check.

**Caveat:** the OKF check uses `filecmp.dircmp` not a content-aware diff, so this would require changing the comparison to be content-normalized rather than byte-exact.

### Option D — Separate OKF check into a warn-only step (recommended)
Like the `badges_changed` signal (warn-only per CLAUDE.md), make OKF drift a `::warning::` that does not fail CI. Only schema/structural OKF changes (new fields, new file names) should be blocking.

**Why this is recommended:** OKF drift is almost always a cosmetic date change, not a structural divergence. Blocking on it creates churn without adding safety.

## Related

- `_VOLATILE_DATE_PATTERNS` in `scripts/build_docs.py` — already normalizes `generatedAt`/`updatedAt`/`version` in JSON; OKF YAML is not covered
- CLAUDE.md "Badge drift is warn-only" precedent — same pattern applied to badges already
- PR #939 — immediate trigger for this issue

mbtiongson1:
Closes #947.

Applies Option D: the same warn-only pattern as badges. `okf_bundle_changed` is reported as a `::warning::` annotation but no longer contributes to the `--check` exit code. Unrelated PRs stop failing CI when `docs/okf/` drifts on the runner.

## Changes
- Add OKF warn-only block (after badges warn-only block) in `scripts/build_docs.py`
- Comment out `okf_bundle_changed` from `changed` flag (same as `badges_changed` precedent)

## Precedent
Same pattern as `badges_changed` (codified 2026-06-24). OKF drift is almost always a cosmetic date change, not a structural divergence.


mbtiongson1:
**Token spend (2026-07-05):** ~$4 total across session (2× scout-haiku, 1× planner, 2× sonnet-worker). Covers both PRs #948 and #949.

PR #1016, opened 2026-07-08
fix(gaia install): tree/ URL misrouting and nested suite failure counting

### LINKED ISSUE #1011: [gaia install] tree/ GitHub URLs are silently misrouted — not parsed, treated as unrecognised URL
## Steps to reproduce
```bash
# If someone registers a skill with a tree/ URL (common GitHub copy-paste mistake):
# links.github: https://github.com/garrytan/gstack/tree/main/review
gaia skills install garrytan/review  # hypothetical with tree/ URL
```

## Expected
Either: parse tree/ as equivalent to blob/ directory, OR reject with a clear error.

## Actual
_parse_github_url() has NO pattern for tree/ URLs.
The blob_match regex fails.
The repo_match regex also fails (because the URL has path segments past owner/repo).
The function returns (url_as_is, None, '') — the original URL is used as the git clone target.
git clone fails or clones the wrong thing.
There is exactly 1 skill in the registry with an unsupported URL format (laravel/upgrade-laravel-v13 uses a GitHub Issues URL), confirming this is a real risk.

## Fix suggestion
Add a tree/ branch to _parse_github_url:
```python
tree_match = re.match(r'https://github\.com/([^/]+)/([^/]+)/tree/([^/]+)/(.*)', url)
if tree_match:
    owner, repo, branch, subpath = tree_match.groups()
    ...
```

## Severity
high — silent failure, no error surfaced to user

### LINKED ISSUE #1015: [gaia install] install_suite only recurses one level — nested suites (suite-of-suites) may partially install
## Steps to reproduce
```bash
# A suite skill whose suiteComponents include another suite (e.g. a meta-suite)
gaia skills install some/meta-suite  # hypothetical nested suite
```

## Expected
Nested suites install fully recursively.

## Actual
install_suite() iterates suiteComponents and calls install_skill() for each.
install_skill() checks suiteComponents on each component and calls install_suite() if present.
This DOES recurse via the visited set — so infinite loops are prevented.
HOWEVER: the success/fail counting in install_suite() only counts the top-level components:
```python
total = len(components) + (1 if root_attempted else 0)
achieved = success_count + (1 if root_attempted and root_ok else 0)
```
If a component is itself a suite with 5 sub-components, and 3 of those fail,
install_suite still counts the parent component as 1 success.
The partial failure in sub-suites is NOT bubbled up to the parent failure tally.

## Impact
Suite install may report 'N/N installed' while silently dropping sub-suite components.

## Severity
medium — only affects nested suites, but the failure is silent

mbtiongson1:
## Problem Statement

Two bugs in `src/gaia_cli/install.py`. First: `_parse_github_url()` has no pattern for GitHub `tree/` URLs — a common curator copy-paste mistake. When a `tree/` URL is registered, the repo_match regex silently matches just the owner/repo portion, drops the subpath entirely, and installs the entire repo root instead of the intended subdirectory. No warning is emitted. Second: `install_suite()` counts each top-level component as a binary success/failure, so when a component is itself a suite (3 exist in the live registry today: `garrytan/gstack`, `mattpocock/skills`, `ruvnet/ruflo`), partial failures inside that nested suite are silently absorbed into a single True return, causing the outer suite to report full success while components are missing.

## Related issues
- Closes #1011
- Closes #1015

## Plan
_To be filled after approval._


mbtiongson1:
Superseded by a proper cli/ branch off main.

PR #1018, opened 2026-07-08
cli(gaia install): fix tree/ URL misrouting and nested suite diagnostic

### LINKED ISSUE #1011: [gaia install] tree/ GitHub URLs are silently misrouted — not parsed, treated as unrecognised URL

### LINKED ISSUE #1015: [gaia install] install_suite only recurses one level — nested suites (suite-of-suites) may partially install

mbtiongson1:
## Problem Statement

Two bugs in `src/gaia_cli/install.py`. First: `_parse_github_url()` had no pattern for GitHub `tree/` URLs — a common curator copy-paste mistake. When a `tree/` URL was registered, the bare-repo regex silently matched just `owner/repo`, dropped the subpath entirely, and installed the entire repo root instead of the intended subdirectory with no warning. Second: `install_suite()` had no way to signal to the operator that a failure in the failed list came from a nested suite rather than a leaf — the bool propagation was already correct, but the diagnostic was opaque.

## Related issues
- Closes #1011
- Closes #1015

## Changes
- `_parse_github_url()`: new `tree/` branch inserted before bare-repo fallback; path is used as subpath directly (always a directory, no dirname step needed)
- `install_suite()`: resolves each component's meta to detect nested suites before installing; annotates nested-suite failures as `comp_id (nested suite — see above)` and prints a hint line pointing operators up the log
- 4 new `_parse_github_url` test cases (tree/ subpath, tree/ root, trailing slash, deep path)
- 1 new nested-suite diagnostic test

## Plan

1. `_parse_github_url()` — add `tree/` URL pattern before bare-repo fallback
2. `_parse_github_url()` — handle `tree/` root + trailing-slash variants
3. `tests/test_install.py` — parametrized tree/ URL cases
4. `install_suite()` — annotate nested-suite failures in the `failed` list
5. `install_suite()` — print nested-suite hint line in the failure summary
6. `tests/test_suite_install.py` — nested-suite diagnostic test

### DIFF
--- a/src/gaia_cli/install.py
+++ b/src/gaia_cli/install.py
@@ -83,6 +83,7 @@ def _parse_github_url(url: str) -> tuple[str, str, str]:
     Examples:
     - https://github.com/owner/repo -> (https://github.com/owner/repo.git, None, "")
     - https://github.com/owner/repo/blob/main/path/to/skill.md -> (https://github.com/owner/repo.git, main, path/to)
+    - https://github.com/owner/repo/tree/main/path/to/skill -> (https://github.com/owner/repo.git, main, path/to/skill)
     """
     url = url.rstrip("/")
     # Pattern for blob URLs: https://github.com/owner/repo/blob/branch/path
@@ -95,6 +96,17 @@ def _parse_github_url(url: str) -> tuple[str, str, str]:
             subpath = os.path.dirname(path)
         else:
             subpath = path
+        return repo_url, branch, subpath
+
+    # Pattern for tree URLs: https://github.com/owner/repo/tree/branch[/path]
+    # tree/ paths always refer to directories — use the path verbatim (no dirname step).
+    # This is a common curator copy-paste mistake; handle it correctly rather than
+    # silently falling through to the bare-repo pattern and dropping the subpath.
+    tree_match = re.match(r"https://github\.com/([^/]+)/([^/]+)/tree/([^/]+)(.*)", url)
+    if tree_match:
+        owner, repo, branch, path = tree_match.groups()
+        repo_url = f"https://github.com/{owner}/{repo}.git"
+        subpath = path.lstrip("/")  # strip leading slash; empty string = repo root
         return repo_url, branch, subpath
 
     # Pattern for base repo: https://github.com/owner/repo
@@ -312,11 +324,22 @@ def install_suite(suite_id: str, registry_path: str, visited: set[str] | None = 
     print(f"\nInstalling suite: {sid} ({len(components)} components)...")
     success_count = 0
     failed: list[str] = []
+    has_nested_suite_failure = False
     for comp_id in components:
+        # Check if the component is itself a suite so we can annotate failures clearly.
+        _, comp_meta = resolve_named_skill_reference(comp_id, registry_path)
+        comp_is_suite = bool(
+            comp_meta and comp_meta.get("suiteComponents")
+        ) if comp_meta else False
+
         if install_skill(comp_id, registry_path, visited, location=location):
             success_count += 1
         else:
-            failed.append(comp_id)
+            if comp_is_suite:
+                failed.append(f"{comp_id} (nested suite — see above)")
+                has_nested_suite_failure = True
+            else:
+                failed.append(comp_id)
 
     # Install the suite root itself if it has its own github source.
     # Use _install_single directly to bypass suite-component detection and
@@ -338,6 +361,11 @@ def install_suite(suite_id: str, registry_path: str, visited: set[str] | None = 
             f"Failed: {', '.join(failed)}",
             file=sys.stderr,
         )
+        if has_nested_suite_failure:
+            print(
+                "  (nested-suite failures listed their own component summaries above.)",
+                file=sys.stderr,
+            )
         return False
 
     print(f"\n✓ Suite {sid} complete: {achieved}/{total} component(s) installed.")
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -75,6 +75,36 @@ class TestInstallInfra:
         assert branch is None
         assert subpath == ""
 
+    def test_parse_github_url_tree_with_subpath(self):
+        """tree/ URL extracts the correct subpath rather than installing the repo root."""
+        url = "https://github.com/garrytan/gstack/tree/main/review"
+        repo, branch, subpath = _parse_github_url(url)
+        assert repo == "https://github.com/garrytan/gstack.git"
+        assert branch == "main"
+        assert subpath == "review"
+
+    def test_parse_github_url_tree_root_only(self):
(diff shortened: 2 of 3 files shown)


mbtiongson1:
## Sandbox Review

> Branch: `cli/gaia-install-findings` · Reviewer: automated sandbox (pi/sonnet-4-5) · 2026-07-08

---

### RED tests (would have failed before fix)

| Description | Covered by test | Notes |
|---|---|---|
| `tree/` URL (`/tree/main/review`) was **silently falling through** to the bare-repo pattern, stripping the subpath and symlinking the whole repo root instead of the `review/` dir | `test_parse_github_url_tree_with_subpath` | Pre-fix: subpath would be `""`, causing the wrong cache dir to be symlinked |
| `tree/main` (no subpath) previously produced `subpath=""` **through the wrong code path** — now explicitly handled | `test_parse_github_url_tree_root_only` | Semantically correct before but untested; now path is explicit |
| `/tree/main/` trailing-slash URL would have produced `subpath="/"` without the lstrip guard | `test_parse_github_url_tree_trailing_slash_stripped` | Outer `url.rstrip("/")` strips trailing slash before regex; group 4 = `""`, lstrip is no-op → `""` ✓ |
| Nested-suite failure in outer summary reported only the suite ID with no annotation, making it impossible to distinguish a plain-leaf failure from a nested-suite failure | `test_nested_suite_failure_annotated_in_outer_summary` | Before fix: `failed.append(comp_id)` for all failures; no `"nested suite — see above"` label |

---

### GREEN tests (pass with fix)

| Description | Result | Notes |
|---|---|---|
| 44 tests total across both files | ✅ PASS | `44 passed, 5 skipped in 1.52s` — skips are `TestRealRegistryIntegrity` (no `registry/named-skills.json` in worktree; expected in this environment) |
| `test_parse_github_url_tree_with_subpath` | ✅ | subpath = `review` |
| `test_parse_github_url_tree_root_only` | ✅ | subpath = `""` |
| `test_parse_github_url_tree_trailing_slash_stripped` | ✅ | with/without trailing slash produce identical tuples |
| `test_parse_github_url_tree_deep_subpath` | ✅ | `skills/foo/bar` preserved verbatim |
| `test_nested_suite_failure_annotated_in_outer_summary` | ✅ | stderr contains `nested suite` and `see above`; manifest contains `testuser/good-leaf` |
| All pre-existing install / suite / uninstall / list tests | ✅ | No regressions |

---

### Edge-case analysis

**Task 4 — `tree_match` regex subpath handling:**

Manually verified with `_parse_github_url()` in the worktree Python env:

| Input URL | branch | subpath | Correct? |
|---|---|---|---|
| `.../tree/main/review` | `main` | `review` | ✅ |
| `.../tree/main` (no path) | `main` | `""` | ✅ |
| `.../tree/main/` (trailing slash) | `main` | `""` | ✅ (outer `url.rstrip("/")` strips the slash before the regex sees it, so group 4 captures `""` and `lstrip` is a no-op) |
| `.../tree/main/skills/foo/bar` | `main` | `skills/foo/bar` | ✅ |

**Task 5 — pre-fetch `resolve_named_skill_reference` in `install_suite` loop:**

(a) **No visited-set corruption.** `resolve_named_skill_reference` accepts only `(skill_ref, registry_path)` — it has no `visited` parameter and performs zero mutations. It calls `list_available()`, which is a pure read (JSON file load + sort). The `visited` set is never passed to it, cannot be mutated by it, and the add/guard in `install_skill` fires exactly once per `comp_id` as before.

(b) **comp_meta correctly discarded.** The returned `comp_meta` dict is used only to read `comp_meta.get("suiteComponents")` and derive the boolean `comp_is_suite`. The variable is not forwarded to `install_skill` — `install_skill` resolves the same skill independently via its own `resolve_named_skill_reference` call. This means each component is resolved twice per install pass. That is a minor performance overhead (O(N·R) where R = registry size) but carries no correctness risk given that `list_available` is read-only and idempotent.

One low-priority note: for registries with thousands of skills and large suites (e.g. `mattpocock/skills`), the double-resolve adds R·K extra dict iterations where K = component count. Acceptable at current registry scale; could be addressed later by passing the pre-resolved meta to `install_skill` via an optional kwarg, but that is out of scope for this PR.

---

### Issues found

- [ ] **Performance (non-blocking):** `resolve_named_skill_reference` is called twice per component (once in the new pre-fetch, once inside `install_skill`). At current registry scale (<500 skills) this is ~1 ms extra per component and does not block correctness. Worth a follow-up ticket if registry grows to 2K+ entries. — `src/gaia_cli/install.py:330`

---

### Verdict: APPROVE ✅

Both changed surfaces (`_parse_github_url` tree-URL handling and `install_suite` nested-suite annotation) are correct, covered by explicit tests, and pass clean. The edge cases requested in the review spec all behave as expected. The only finding is a non-blocking performance note that does not warrant changes in this PR.

PR #1024, opened 2026-07-08
infra: extract scripts/lib/ shared primitives (PR 2/7)

nova-gaia:
## Summary

Introduces `scripts/lib/` — the shared infrastructure primitive package described in upstream-watcher design §9 (Shared plumbing). No behavior changes to any existing script. `stargazerHeartbeat.py` is untouched (zero-byte diff against main).

Design reference: `docs/agents/upstream-watcher.md` [§9 Shared plumbing (`scripts/lib/`)](https://github.com/gaia-research/gaia-skill-tree/blob/main/docs/agents/upstream-watcher.md#shared-plumbing-scriptslib)

Part of the 7-PR rollout table in design §14. Follows PR 1 (labels + template). PR 3 will rewrite `stargazerHeartbeat.py` to import from this package.

---

## Files created

| File | Lines | Description |
|---|---|---|
| `scripts/lib/__init__.py` | 14 | Package marker + `__version__` |
| `scripts/lib/frontmatter.py` | ~295 | `split_frontmatter`, `load_yaml_simple`, `update_list_item_in_frontmatter`, `upsert_top_level_block` |
| `scripts/lib/github_api.py` | ~175 | `parse_owner_repo`, `fetch_json`, `head_check` |
| `scripts/lib/named_iterator.py` | ~60 | `iter_named_skills(root)` |
| `scripts/lib/tests/__init__.py` | 5 | Test package marker |
| `scripts/lib/tests/test_lib.py` | ~500 | 43 unit tests — all passing locally |

---

### `scripts/lib/frontmatter.py`

- **`split_frontmatter(text)`** — ported from `stargazerHeartbeat._split_frontmatter`, renamed to public API. Splits raw file text into `(fence, fm_content, body)`.
- **`load_yaml_simple(text)`** — thin pyyaml wrapper. Ported from heartbeat.
- **`update_list_item_in_frontmatter(text, list_key, row_index, field_updates)`** — generalised form of heartbeat's `_update_evidence_row_in_text`. Works for any list key (evidence, timeline, etc.). The heartbeat use-case (`list_key="evidence"`, `field_updates={"stars": N, "updatedAt": ...\}`) passes through unchanged.
- **`upsert_top_level_block(text, block_key, block_value)`** — new helper for writing/merging an entire top-level frontmatter block (e.g. `upstream:`). Deterministic key ordering. Merge-preserves unrelated sub-keys.

### `scripts/lib/github_api.py`

[...]

PR #1025, opened 2026-07-08
infra: refactor stargazerHeartbeat.py to use scripts/lib/ (PR 3/7)

nova-gaia:
## Summary

Refactor `scripts/stargazerHeartbeat.py` to import from `scripts/lib/` (introduced in PR 2/7, merged as #1024). Deletes all local copies of shared primitives. Zero behavior change.

### `scripts/stargazerHeartbeat.py` (410 → 272 lines, -138)

| Deleted local primitive | Replaced with |
|---|---|
| `_FM_RE`, `_split_frontmatter` | `split_frontmatter` from `scripts.lib.frontmatter` |
| `_load_yaml_simple` | `load_yaml_simple` from `scripts.lib.frontmatter` |
| `_GH_RE`, `parse_owner_repo` | `parse_owner_repo` from `scripts.lib.github_api` |
| `_CACHE`, `fetch_stars` | thin `_fetch_stars()` wrapper calling `fetch_json` from `scripts.lib.github_api`; extracts `data["stargazers_count"]` at call site |
| `_update_evidence_row_in_text` call site | `update_list_item_in_frontmatter(text, "evidence", idx, {"stars": N, "updatedAt": today})` from `scripts.lib.frontmatter` |

**Additional fix (latent bug, not a behavior change):** the original `fm.get("id", path.relative_to(NAMED_DIR).with_suffix(""))` always evaluated the fallback expression eagerly (Python `dict.get` evaluates the default before calling). This caused a `ValueError` when `process_file` was called on a file outside `NAMED_DIR` (e.g. in tests). Fixed to use an explicit `if "id" in fm` check with a `try/except ValueError` fallback. Zero behavior change in production (all real named skill files have `id` in frontmatter AND live under `NAMED_DIR`).

**sys.path bootstrap added:** `scripts/` has no `__init__.py` (setuptools only packages `src/gaia_cli`). The CI workflow runs `python scripts/stargazerHeartbeat.py --apply` from the repo root after `pip install -e ".[dev]"`, which adds `." to `pythonpath` via `pyproject.toml`. The bootstrap (`sys.path.insert(0, repo_root)`) covers any invocation context where the repo root isn't already on `sys.path`.

### `scripts/lib/tests/test_heartbeat_regression.py` (new, 284 lines)

12 regression tests across 5 classes:
- `TestHeartbeatDryRun` — result row shape, no file writes in dry-run mode
- `TestHeartbeatApplyMode` — stars written, updatedAt set to today, **golden output byte-matches `update_list_item_in_frontmatter` direct call**, non-star rows untouched, body preserved
- `TestHeartbeatApiErrors` — None response → skip_reason=api_error, no file write
- `TestHeartbeatBelowThreshold` — zero delta → no write
- `TestHeartbeatNoGithubUrl` — non-GitHub source URL → skip_reason=no_github_url

## Zero behavior change attestation

| Verification | Result |
|---|---|
| All 7 local primitives deleted from heartbeat | ✅ confirmed — zero local copies remain |
| `_needs_update` threshold logic unchanged | ✅ untouched — same two conditions, same values |
| CLI surface unchanged (`--dry-run`, `--apply`, exit codes, log format) | ✅ no changes to `main()` or output format |
| Write cadence, throttle, error handling unchanged | ✅ delegated to `fetch_json` which carries identical 100 ms throttle and error semantics |
| `stargazer-heartbeat.yml` workflow unchanged | ✅ not touched |
| `python scripts/stargazerHeartbeat.py --apply` still works from repo root | ✅ verified via dry-run (verification step 2) |

**Golden output proof:** `test_apply_golden_output_matches_update_list_item` in the new test file calls `heartbeat.process_file(fixture_path, apply=True)` and `update_list_item_in_frontmatter(FIXTURE_TEXT, "evidence", 1, {"stars": 220000, "updatedAt": today})` independently, then asserts byte equality. This proves the delegation is exact.

## Verification step outputs

**Step 1 — all imports succeed:**
```
$ python -c "import scripts.lib.frontmatter; import scripts.lib.github_api; import scripts.lib.named_iterator; import scripts.stargazerHeartbeat; print('All imports OK')"
All imports OK
```

**Step 2 — dry-run completes without errors (first 20 lines):**
```
stargazerHeartbeat — mode: DRY-RUN
Scanning .../registry/named ...

skill_id                                   old_stars  new_stars    delta  updated
----------------------------------------------------------------------------------
addy-osmani/agent-skills                       68564      73353    +4789  would
addy-osmani/code-review-and-quality            68564      73353    +4789  would
addy-osmani/code-simplification                68564      73353    +4789  would
addy-osmani/incremental-implementation         68564      73353    +4789  would
addy-osmani/performance-optimization           68291      73353    +5062  would
addy-osmani/planning-and-task-breakdown        68564      73353    +4789  would
addy-osmani/shipping-and-launch                68564      73353    +4789  would
addy-osmani/spec-driven-development            68564      73353    +4789  would
garrytan/benchmark                            118552     120474    +1922  would
garrytan/canary                               118552     120474    +1922  would
...
```
(No ImportError, TypeError, or AttributeError — API calls succeeded because GH_TOKEN was available in shell)

**Step 3 — pytest 55/55 passing:**
```
55 passed in 2.25s
```

## Judgment calls

1. **`iter_named_skills` not used in the main loop.** The heartbeat's `main()` uses `NAMED_DIR.rglob("*.md")` and calls `process_file(path, apply=args.apply)`. `iter_named_skills` would yield `(path, fm)` pairs, but `process_file` re-reads the file itself (it needs the raw text for rewriting). Replacing the loop would require either passing the pre-read `text` into `process_file` (signature change) or reading the file twice (wasteful). Per the spec: "either replace with the iterator or leave as-is if the shape doesn't match cleanly. Prioritize zero-behavior-change over aesthetic cleanup." Left as-is.

2. **`_fetch_stars` is a 4-line local wrapper** rather than inline call-site expansion, because the call site is inside a loop over evidence rows (multiple stars fetches per file are possible). A named wrapper is clearer than `data = fetch_json(...); new_stars = data.get("stargazers_count") if data else None` repeated inline.

3. **sys.path bootstrap location.** Placed between `from pathlib import Path` and the `scripts.lib` imports, executed unconditionally. Alternative was conftest.py injection, but that would only work for pytest — the workflow runs the script directly.

## Branch scope

`infra/upstream-watcher-heartbeat-refactor` — touches `scripts/` and `scripts/lib/tests/`, both in the `infra/` allowlist.

## Deferred surfaces

None.

## Entrypoints

Not applicable — no user-facing page added.

### DIFF
--- a/scripts/lib/tests/test_heartbeat_regression.py
+++ b/scripts/lib/tests/test_heartbeat_regression.py
@@ -1,0 +1,284 @@
+"""
+scripts/lib/tests/test_heartbeat_regression.py — regression test for stargazerHeartbeat.py.
+
+Verifies that the refactored heartbeat (PR 3/7) produces byte-for-byte identical
+output to what the original ``_update_evidence_row_in_text`` logic would have
+written, by running ``process_file`` against a fixture file and asserting the
+golden output.
+
+Test strategy
+-------------
+We exercise the heartbeat's ``process_file`` in both dry-run (``apply=False``)
+and apply (``apply=True``) modes against an in-memory fixture file.  ``fetch_json``
+is mocked to return a canned API response so no live network calls are made.
+
+The golden output is computed from the known fixture contents + the known mock
+star count, using ``update_list_item_in_frontmatter`` directly (the same
+function that the refactored heartbeat now delegates to).  This proves:
+
+1. The heartbeat correctly delegates to ``update_list_item_in_frontmatter``.
+2. The delegation is equivalent to the original ``_update_evidence_row_in_text``
+   (which was already tested in ``test_lib.py``).
+3. The ``_fetch_stars`` wrapper correctly extracts ``stargazers_count`` from
+   the ``fetch_json`` response.
+4. The ``_needs_update`` threshold still gates writes correctly.
+
+Note on fixture data
+--------------------
+The fixture is a minimal synthetic version of ``mattpocock/diagnose.md``
+(a real named skill with a ``github-stars-own`` evidence row).  It is NOT
+a copy of the live file — the live file changes as stars update, which would
+make the test fragile.  The fixture has a known star count (100_000) and the
+mock API returns 220_000, which exceeds the 5% threshold so a write is
+expected.
+"""
+
+from __future__ import annotations
+
+import textwrap
+from pathlib import Path
+from datetime import date
+from unittest.mock import patch
+
+import pytest
+
+import sys
+
+_REPO_ROOT = Path(__file__).resolve().parent.parent.parent.parent
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+from scripts.lib.frontmatter import update_list_item_in_frontmatter
+import scripts.stargazerHeartbeat as heartbeat
+
+# ---------------------------------------------------------------------------
+# Fixture text — synthetic, based on mattpocock/diagnose.md structure
+# ---------------------------------------------------------------------------
+
+_FIXTURE_STARS_STORED = 100_000
+_MOCK_API_STARS = 220_000  # > 5% delta AND > 100 absolute → should update
+
+FIXTURE_TEXT = textwrap.dedent(f"""\
+    ---
+    id: mattpocock/diagnose
+    name: Diagnose
+    contributor: mattpocock
+    status: named
+    level: 3★
+    links:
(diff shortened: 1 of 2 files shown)


nova-gaia:
**Token spend — PR 3/7 session (2026-07-09)**

Sonnet 4.5: ~35k in, ~12k out. ~$0.35

Issue #868, opened 2026-06-27
feat(leaderboard): Trust Leaderboard redesign — SVG bar chart, dark atmosphere, interactive


mbtiongson1:
## Goal

Redesign `docs/trust/leaderboard/index.html` from a table-based layout into a data-forward SVG vertical bar chart visualization inspired by artificialanalysis.ai.

## Depends on
Sprint B integration branch `dev/sprint-b2-trending` (PR #863)

## Deliverables
- [x] `docs/trust/leaderboard/index.html` — redesigned with SVG chart containers, dark-page body class, tooltip div
- [x] `docs/trust/leaderboard/leaderboard.css` — dark atmosphere (`oklch(0.09 0.025 250)`), SVG bar animations, pill filters, fixed tooltip
- [x] `docs/trust/leaderboard/leaderboard.js` — SVG vertical column charts, stacked bars for suites, grade-colored bars, mouse-tracking tooltip, progressive loading (top 40 + show more), filter/sort controls
- [ ] Visual review + surgical iteration pass (nitpicks pending — session 25 paused here)
- [ ] `--rank-N-rgb` tokens added to `scripts/generateCssTokens.py` (JS currently hardcodes RGB fallbacks)
- [ ] Wire trending delta column once `/api/v1/trending/7d.json` is available

## Design direction
- Trust Magnitude = primary index (Y axis). Bars go UP.
- Three swim lanes: Suites (stacked by rank tier) · Named Skills (grade-colored bars) · In Registry (compact list, no chart)
- Dark isolation: leaderboard gets its own background atmosphere, distinct from rest of site
- Interactive: hover tooltip with TM, grade, level, component count. Touch-first mobile.
- Filter pills by grade (S/A/B/C). Sort by TM/Trending/Grade.
- Design tokens only — zero hex in CSS (CI guard compliant)

## Branch
`dev/leaderboard-redesign` → `dev/sprint-b2-trending` (PR #867)

## Size
M (implementation done, iteration pending)

mbtiongson1:
Shipped in #1395 and released in v7.3.8.

The Trust Leaderboard redesign/polish landed on `main` as part of the issue-backlog integration, including the final intake CTA and local/founder review fixes. Closing as complete.

Issue #346, opened 2026-05-20
plaques in canon flowchart tree view in Named Skills Explorer still "flickering"


mbtiongson1:
Here is a sample. When hovering, the plaque moves a bit to the right. z-index is also wrong here-- the nodes overlap it. Need help!

<img width="655" height="378" alt="Image" src="https://github.com/user-attachments/assets/05847656-cf00-4739-898e-05bf3934d2db" />

mbtiongson1:
Resolved in docs/css/styles.css (L7478 overflow:visible ancestors, L11445 plaque z-index:60, L11460+ transform pinned while open). Addresses 'moves right on hover' + 'nodes overlap it'. Closed on the post-v7.1.1 reconciliation pass.

Issue #235, opened 2026-05-12
design


mbtiongson1:
Search skills in svg does not work with other names, only with title. It should work with the name and the slash skill.

nova-gaia:
Triage update: this appears addressed in the interactive graph. `docs/js/skill-graph.js` `_searchMatches()` now supports name and slash-form search, not just title: slash-prefix queries (`/foo`) match skill id, name, and the slash slug (`slugSlash`), and free-text queries match id/name/description/title/namedId. Note that search only exists in the interactive graph — the static `gaia.svg` (rendered by `graph.py::render_svg`) only carries `<title>` tooltips and has no search layer.

If the interactive-graph behavior now covers your case, this can be closed. If you specifically meant search over the static SVG export, please confirm and we can re-scope.

nova-gaia:
Closing as **already implemented**. The graph search now matches by name and slash-name, not just title.

Evidence — `docs/js/skill-graph.js::_searchMatches` (L901–919):
- **Free-text mode** matches across `id`, `name`, `description`, `title`, **and the full `contributor/slug` slash-name** (`namedId.toLowerCase().includes(searchTerm)`, L918).
- **`/slash` prefix mode** (L911) matches skill id/name + the slug half of the slash-name.
- **`@handle` prefix mode** (L914) matches contributor handles.

So typing a nickname, a `/slash-skill`, a `@handle`, or free text all resolve. This directly covers the ask ("should work with the name and the slash skill").

Verified during the triage sprint (Wave 2). No code change required.

PR #1106, opened 2026-07-09
fix(design): flowchart popover flicker (#346) + --rank-N-rgb tokens (#868)

nova-gaia:
## Wave 2 — frontend bundle → `dev/triage-sprint-2026-07`

Part of the triage sprint. Staged into the integration branch, **not** main.

### #346 — Flowchart / DAG popover flicker (long-lived)
Triangulated to **two independent root causes**, both fixed in `docs/css/styles.css`:

**1. Transform-flip.** Popover cards are centered with `transform: translateX(-50%)`. When the cursor crosses onto the card, the card's own `:hover` transform fires and overrides the centering:
- `.plaque--mini:hover { transform: none !important }` (L9886) for mini/DAG cards
- `.flow-node:hover { transform: translateY(-3px) }` (L7402) for upgrade-path cards

Either shifts the card, the cursor falls off, and hover thrashes → flicker. The prior pin `.git-node > .plaque:hover` lacked `!important` (lost to the `!important` mini rule) and didn't cover `.flow-node` at all. **Fix:** pin `transform: translateX(-50%) !important` for every popover card type while open, scoped under `.git-node:hover` so standalone flow-nodes keep their lift.

**2. Stacking-context trap.** The rank rows (`.ns-dag-rank` / `.se-flowchart-row` / `.git-rank-row`) each set `z-index: 2` (L10521), creating a stacking context that caged the hovered node's `z-index: 30` **inside its own row** — so the next rank row's nodes painted over the downward-opening popover. **Fix:** `:has(.git-node:hover) { z-index: 30 }` lifts the row that actually contains the hovered node above its sibling rows. (`:has()` is already an established pattern in this file, L417/420.)

### #868 — `--rank-N-rgb` design tokens
The Trust Leaderboard JS (`docs/trust/leaderboard/leaderboard.js`) read `tok('--rank-N-rgb')` but `tokens.css` never emitted those triplets, so it always fell back to hardcoded RGB values.

**Fix:** added the `--rank-<N>-rgb` line to `_emit_rank_block` in `scripts/generateCssTokens.py` (the tier block already had `--tier-<name>-rgb`; ranks were the gap). Regenerated `docs/css/tokens.css` — pure additions, values match the JS fallbacks exactly. Fallbacks kept as a defensive floor for pre-token-load, comment updated. No version-stamp churn (#807).

### #235 — SVG/graph search by name + slash-name
**Verified already resolved** — closing separately with evidence, no code change here. `docs/js/skill-graph.js::_searchMatches` (L901) already matches `id`, `name`, `description`, `title`, **and** the full `contributor/slug` `namedId`, plus dedicated `/slash` and `@handle` prefix modes (L911-918). Directly satisfies "should work with the name and the slash skill."

## Tests
- `tests/test_tui_tokens.py`: 9 passed (token generator).
- CSS: braces balanced (1726/1726), zero hex literals in the #346 region.
- Guard A (hardcoded-colors) clean on all touched `docs/css` + `docs/js` files.

## Scope note
`design/` scope is `docs/` + `*.md`; this PR also touches `scripts/generateCssTokens.py` (the source of the `tokens.css` regen for #868). Needs **`skip-scope-check`** — the generator change and its emitted `tokens.css` must move together.

## Entrypoints
N/A — no new page or nav surface. #346 fixes an existing interaction on the Named-skills DAG + skill-explorer upgrade path; #868 wires existing leaderboard bars to canonical tokens.

### DIFF
--- a/docs/css/styles.css
+++ b/docs/css/styles.css
@@ -10326,13 +10326,34 @@ button.nav-graph-trigger:hover {
   z-index: 30;
 }
 
-/* The popover card is itself a .plaque, so the global .plaque:hover
-   lift (translateY(-2px)) competes with the translateX(-50%) centering
-   when the cursor crosses onto the card — the card shifts sideways, the
-   cursor leaves it, and the hover state thrashes. Pin the transform for
-   the card while it is open. */
+/* Root cause 2 — the rank rows (.ns-dag-rank / .se-flowchart-row /
+   .git-rank-row) each establish a stacking context at z-index: 2 (see the
+   rank-row rule below). A hovered node lifted to z-index: 30 is still trapped
+   *inside* its own row's context, so its downward popover was painted over by
+   the NEXT row's nodes. Lift the row that actually contains the hovered node
+   above its sibling rows. */
+.ns-dag-rank:has(.git-node:hover),
+.se-flowchart-row:has(.git-node:hover),
+.git-rank-row:has(.git-node:hover) {
+  z-index: 30;
+}
+
+/* Root cause 1 — the popover card is positioned with transform:
+   translateX(-50%). When the cursor crosses onto the card, the card's own
+   :hover transform fires and overrides the centering:
+     • .plaque--mini:hover { transform: none !important }  (mini/DAG cards)
+     • .flow-node:hover    { transform: translateY(-3px) } (upgrade-path cards)
+   Either one shifts the card sideways/up, the cursor leaves it, and the hover
+   state thrashes → flicker. Pin the centering transform for every popover card
+   type while it is open. !important + the descendant chain beat the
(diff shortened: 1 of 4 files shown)


nova-gaia:
**Token spend** (Wave 2 frontend bundle)

2026-07-10 Opus 4.8 Extra High: ~120k in, ~15k out. ~$5

Covers: #346 two-root-cause flicker fix, #868 --rank-N-rgb token emission + tokens.css regen, #235 verification.

mbtiongson1 (approved):

mbtiongson1:
Approved, could be problematic with the current ygdrasil implementation, will conflict. Lets see how we can cleanly stage this

PR #1111, opened 2026-07-09
docs: codify sprint completeness rule (no follow-ups) + close #611

nova-gaia:
## What

Two things, both closing out the 2026-07 triage sprint so it leaves **zero** follow-ups:

1. **`CLAUDE.md` — new "Sprint Completeness — no follow-ups" section.** Codifies the operator directive: every sprint ships complete; spillover is scoped during staging as an extra PR *within* the sprint, never deferred as a follow-up issue. Genuinely new out-of-scope discoveries are still normal backlog hygiene — what's forbidden is treating a sprint's own remainder as "future work."

2. **Closes #611** (`gaia push --update` incremental pushes). The feature shipped in PR #1107 (merged to main via #1109): \`skill_identity_key\` + \`skill_fingerprint\` identity model, \`classify_against_pending\` NEW/CHANGED/UNCHANGED classifier, and the issue-body identity-block round-trip. Covered by 23 passing tests in \`tests/test_push_incremental.py\`.

## Why in this sprint

Per the new rule this PR adds: the trending CI false positive (#1108, fixed in the sibling PR) and this #611 closure are both direct consequences of the triage sprint's own work, so they're resolved *inside* the sprint rather than deferred.

## Entrypoints
N/A — docs/policy change only, no new user-facing page or section.

PR #1110, opened 2026-07-09
fix(ci): make trending drift warn-only in gaia dev docs --check (#1108)

nova-gaia:
## Problem

\`docs/api/v1/trending/{7d,30d}.json\` + \`feed.xml\` are **rolling time-window** artifacts. Their bytes recompute the moment the UTC day advances past the last release regen — independent of any PR's content. Because \`trending_changed\` fed the \`gaia dev docs --check\` exit code, every PR that merely triggers \`validate.yml\` (touches \`tests/**\`, \`scripts/**\`, etc.) failed **Schema + DAG + Integrity Checks** on trending drift it did not cause, once run on a later UTC day than the last regen.

### Evidence
- PR #1107 (\`cli/\` branch, 6 files, zero docs artifacts) failed on \`7d.json\` + \`feed.xml\` on 2026-07-10; its committed trending files were byte-identical to base.
- PR #1105 (same shape) passed the identical check on 2026-07-09 — the UTC day the trending files were baked at v6.3.15.
- Only variable between pass and fail: the wall clock.

## Fix

Option 1 from #1108 (warn-only, the low-risk choice matching existing precedent):

- Remove \`trending_changed\` from the \`changed\` aggregation that feeds the \`--check\` exit code.
- Emit a \`::warning::\` instead, mirroring the okf/badges warn-only blocks directly above it and the ISO-week rollover skip in \`build_content_engine\`.

The production cron still regenerates the real windows; the drift stays visible in \`--check\` diff output for inspection. No change to the trending engine's output or to \`validate.yml\` trigger paths.

## Testing

- \`python -c "import ast; ast.parse(...)"\` — syntax OK.
- Full suite run; remaining failures (\`test_scanner\`, \`test_push\`, \`test_workspace_mode\`, \`test_timelines\`, etc.) reproduce identically on clean \`main\` — pre-existing Windows-environment failures (\`ValueError: I/O operation on closed file\`, path-casing), not regressions. This change only removes a boolean from a CI-gate aggregation.

## Entrypoints
N/A — no new user-facing page or section; CI-gating change only.

Closes #1108.

PR #1112, opened 2026-07-09
fix(ci): fix 100%-failing Release workflow (regenerate Class P index before validators)

nova-gaia:
## Problem — Release workflow has a 100% failure rate

\`release.yml\` (fires on \`v*\` tag push) gates the GitHub Release on a \`validate-graph\` job. That job installed only \`jsonschema\`+\`PyYAML\` and ran the validators **directly**:

\`\`\`
- name: Install dependencies
  run: python -m pip install jsonschema==4.26.0 PyYAML==6.0.1
- name: Validate redaction invariant …
  run: python scripts/validate_redaction.py
\`\`\`

But \`validate_redaction.py\` reads **\`registry/named-skills.json\`** as its source of truth — a **gitignored Class P artifact** that does not exist on a fresh CI checkout. So the step dies every time:

\`\`\`
redaction: cannot find …/registry/named-skills.json
##[error]Process completed with exit code 1.
\`\`\`

The \`release\` job (\`needs: [test-python, test-mcp, validate-graph]\`) then never runs. Confirmed failing on **v6.0.0, v5.1.0, v4.4.0, v4.2.1** (Python Tests ✓, MCP Tests ✓, Validate Graph ✗ every time). Every GitHub Release since has been produced by \`sync-artifacts.yml\`'s fallback \`gh release create\` path — which is why tags accumulated as canary pre-releases and Latest is still stuck at v6.0.0.

## Fix (yml-only)

Mirror what \`validate.yml\` already does correctly on every PR — install the package and regenerate the intermediate registry artifacts before validating:

- \`pip install -e ".[dev]"\` (so \`gaia_cli\` is importable by the generators)
- new **Build intermediate registry artifacts** step: \`assemble_gaia.py\` + \`generateNamedIndex.py\` + mirror to \`docs/graph/named/index.json\`

**No change to how Class P/S artifacts are stored.** The index stays gitignored and is regenerated in-CI, byte-identical to \`validate.yml\`'s approach.

## Verification

Reproduced the failure and the fix locally (Windows checkout):
- Index **absent** → \`validate_redaction.py\` **exit 1** (the CI failure).
- Run \`assemble_gaia.py\` + \`generateNamedIndex.py\`, index **present** → \`validate.py\`, \`validate_redaction.py\` **exit 0**.
- YAML syntax validated with \`yaml.safe_load\`.

## Entrypoints
N/A — CI workflow fix only, no user-facing surface.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,20 @@ jobs:
           python-version: '3.12'
 
       - name: Install dependencies
-        run: python -m pip install jsonschema==4.26.0 PyYAML==6.0.1
+        run: python -m pip install -e ".[dev]"
+
+      # registry/named-skills.json (+ the docs/graph mirror) is a gitignored
+      # Class P artifact — absent on a fresh checkout. The validators below read
+      # it as their source of truth, so it MUST be regenerated first, exactly as
+      # validate.yml does on every PR. Without this step validate_redaction.py
+      # dies with "cannot find registry/named-skills.json" and the release job
+      # (needs: validate-graph) never runs — the 100%-failure root cause.
+      - name: Build intermediate registry artifacts
+        run: |
+          python scripts/assemble_gaia.py
+          python scripts/generateNamedIndex.py
+          mkdir -p docs/graph/named
+          cp registry/named-skills.json docs/graph/named/index.json
 
       - name: Validate schema + DAG + evidence
         run: python scripts/validate.py


nova-gaia:
## Session token spend

`2026-07-09/10 Opus 4.8 Extra High: ~340k in, ~48k out. ~$8.75`

Covers: triage follow-up PRs (#1110 trending warn-only, #1111 sprint-completeness rule), release-automation fix (#1112 — repaired 100%-failing `release.yml` validate-graph gate), v6.4.0 cut + promotion to Latest, and manual PyPI publish recovery.

PR #1113, opened 2026-07-10
fix(docs): regenerate stale docs and update cache-busting version to 6.4.0

milim-gaia:
This PR regenerates the stale docs/ files and syncs their cache-busting version query strings with version 6.4.0, fixing the failing Validate Gaia Graph CI check on main.


milim-gaia:
2026-07-10 Gemini 1.5 Pro: ~25k in, ~3k out. ~sh.20

mbtiongson1 (commented):
checking, adding a second fix pr and comparing

nova-gaia:
Closing in favor of #1114 (now merged), which fixes the **root cause** rather than this instance.

#1113 regenerates and commits the stale snapshot (81 files, including cron-owned trending JSON and `layouts_3d.json` float jitter). It goes green, but the identical `Validate Gaia Graph` failure recurs on the **next** release because `gaia dev release` still never re-stamps the version-derived surfaces.

#1114 teaches `gaia dev release --cache-bust-only` to refresh both the HTML `?v=` strings **and** the README version region on every release commit, so no future release drifts. Regenerating the volatile data artifacts is left to the cron/auto-sync flow that owns them.

PR #1114, opened 2026-07-10
fix(ci): fold HTML cache-bust rewrite into gaia dev release

nova-gaia:
## Problem

`validate.yml` fires on every release commit (`packages/cli-npm/**` is in its path filter) and runs `gaia dev docs --check`. The check fails because `gaia dev release` bumps `pyproject.toml`/`package.json` but never updates the `?v=X.Y.Z` strings in the 26 tracked Class S `docs/*.html` files — so the version in those strings lags by one release every time.

Root cause confirmed by the v6.4.0 failure ([run 29053497042](https://github.com/gaia-research/gaia-skill-tree/actions/runs/29053497042)): all 26 HTML files still had `?v=6.3.x` after the v6.4.0 release commit landed.

## Fix (Opus-reviewed)

- **`scripts/build_docs.py` — `--cache-bust-only` mode**: runs only `build_html_cache_busting` in write mode, prints `cache-bust: updated <path>` for each file touched, exits 0. Never regenerates badges, `gaia.json`, or other Class S artifacts — avoids the 2026-06-24 stale-snapshot badge wipe risk associated with full `gaia dev docs`.
- **`src/gaia_cli/impl.py` — `release_command`**: calls `build_docs.py --cache-bust-only` after `bump_versions` and stages the rewritten HTML files into the release commit alongside `pyproject.toml` and the `package.json` files.

Also fixes the **existing v6.4.0 drift** on `main` (all 26 HTML files still had `?v=6.3.x`).

## Why not Option B (skip the check on `[skip-gen]` commits)?

The `?v=X.Y.Z` strings gate real browser asset fetches — they are functional cache-busting, not decoration. Skipping the check would mean a release, the one moment new CSS/JS ships, is exactly when cache-busting silently fails to fire on cached browsers.

## What is NOT touched

- `release.yml` — PyPI fix from PR #1112 is untouched
- `validate.yml` — no changes to the check
- `sync-artifacts.yml` — `[skip-gen]` skip still applies; no badge regen risk

## Entrypoints

N/A — CI and CLI tooling only, no new user-facing surface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ curl https://gaiaskilltree.com/api/v1/leaderboard.json
 **1. CLI
 
 <!-- gaia:version-start -->
-Current Gaia CLI version: `6.3.20`.
+Current Gaia CLI version: `6.4.0`.
 
 ```bash
 curl -fsSL https://gaiaskilltree.com/install.sh | sh
--- a/docs/about.html
+++ b/docs/about.html
@@ -4,7 +4,7 @@
   <meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate">
   <meta http-equiv="Pragma" content="no-cache">
   <meta http-equiv="Expires" content="0">
-  <script>window.GAIA_VERSION = "6.3.20";</script>
+  <script>window.GAIA_VERSION = "6.4.0";</script>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <title>About — Gaia Skill Registry</title>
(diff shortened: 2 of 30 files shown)


nova-gaia:
## Root-cause fix completed (Opus 4.8)

The v6.4.0 `Validate Gaia Graph` failure had **two** version-derived surfaces drifting on every release; this PR originally closed only one:

1. `?v=X.Y.Z` / `GAIA_VERSION` strings in tracked Class S `docs/*.html` — ✅ (original commit)
2. the README version region (`<!-- gaia:version-start --> … end`) — ❌ was still stale

Because `gaia dev docs --check` validates **both**, the release commit stayed red even after the HTML rewrite (README still said `6.3.20`). That's why this PR's own `Schema + DAG + Integrity Checks` was failing.

### What landed in `400d40b17`
- **`build_docs.py`**: new `build_readme_version_only()` — refreshes **only** the README version region (derives solely from `pyproject.toml`, never the gitignored Class P registry snapshot). Wired into `--cache-bust-only` alongside the HTML rewrite; emits `cache-bust: updated README.md` so `release_command` stages it automatically.
- **`impl.py`**: broadened the release-time comment to name both surfaces.
- **README**: closed the existing v6.4.0 drift (`6.3.20` → `6.4.0`).
- **`tests/test_build_docs_cache_bust_only.py`**: 4 tests pinning the contract — version-region-only rewrite, idempotency, missing-file no-op, and a guard that `--cache-bust-only` never runs the full pipeline (asserts `build_badges` is not invoked, protecting against the 2026-06-24 badge-wipe class of failure).

### Verification
- `python scripts/build_docs.py --check` → **rc=0** locally
- `Schema + DAG + Integrity Checks` → **pass** on CI (was failing)
- All 21 related tests + 4 new tests pass
- No data-artifact churn: the commit is exactly 4 files (+168/−9), no `gaia.json`/`skill-trees/` timestamp or float-jitter noise

### Branch scope
`src/gaia_cli/impl.py` sits outside the `infra/` allowlist, but this is one atomic CI fix spanning `src/` + `scripts/` + `docs/` — splitting it across a `cli/` and `infra/` branch would be strictly worse. Applied the documented `skip-scope-check` label per CLAUDE.md.

### Why this over PR #1113
#1113 regenerates-and-commits the stale snapshot (81 files, incl. cron-owned trending JSON + `layouts_3d.json` float jitter). It's green but is a treadmill fix — the identical failure recurs on the **next** release. This PR fixes the mechanism so no future release drifts.

---
**Token spend:** 2026-07-10 Opus 4.8: ~95k in, ~9k out. ~$2.10

PR #1115, opened 2026-07-10
docs(DEV): python -m gaia_cli fallback map for blocked environments

nova-gaia:
## What

Adds **DEV.md §2A — Blocked-Environment Fallback**: a verified 1:1 mapping of every `gaia dev` command to a `python -m gaia_cli dev` alternative, for environments where the `gaia` shim is unavailable (enterprise Windows blocking unknown PATH executables, sandboxes stripping `~/.local/bin`, or worktrees where `gaia` resolves to the wrong copy).

## Why

Agents currently hit a hard wall when the `gaia` binary is blocked, with no documented safe alternative. This gives them one.

## Core finding (mechanical, not per-command)

`pyproject.toml` defines `gaia = "gaia_cli.main:main"`, and `python -m gaia_cli` (`__main__.py`) calls the **same** `main()`. They share one argv/dispatch path:

```
gaia <anything>  ≡  python -m gaia_cli <anything>
```

## What's in the section

- The universal substitution rule + examples.
- **Three env knobs**: `PYTHONPATH=…/src` (worktree source), `GAIA_OPERATOR_OVERRIDE=1` (mutating verbs outside a Verifier context), `PYTHONUTF8=1 PYTHONIOENCODING=utf-8` (Windows — the `build_docs.py` subprocess doesn't inherit the parent's UTF-8 reconfigure).
- **Per-command table** covering all 28 `dev` subcommands: mutating flag, any genuine `scripts/*.py` direct alternative (only `build`, `docs`, `validate --intake`, `test`), and load-bearing caveats.

## Verification

Mapping was adversarially verified — one skeptic agent per command read the actual source. All 28 subcommands confirmed `python -m gaia_cli` as a byte-for-byte substitution (zero failures). Gotchas surfaced and now documented: `dev verify` is double-gated (ignores the override), `dev merge --named` is a no-op, `sync-upstream`'s real flag is `--tag` (docstring's `--version` is stale).

## Scope

Docs-only. Single file, +69 lines, no source/data touched.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -107,6 +107,75 @@ pip install -e ".[docs]"
 
 ---
 
+## 2A. `gaia dev` → `python -m gaia_cli dev` (Blocked-Environment Fallback)
+
+**Use this when the `gaia` shim is unavailable** — e.g. enterprise Windows that blocks unknown executables on `PATH`, sandboxes that strip `~/.local/bin`, or a worktree where `gaia` resolves to the wrong (system-installed) copy.
+
+### The universal rule (1:1, mechanical)
+
+The console script `gaia` is defined in `pyproject.toml` as `gaia = "gaia_cli.main:main"`. The module entry `python -m gaia_cli` runs `src/gaia_cli/__main__.py`, which calls the **same** `gaia_cli.main:main`. They share one argv/dispatch path, so:
+
+```
+gaia <anything>   ≡   python -m gaia_cli <anything>
+```
+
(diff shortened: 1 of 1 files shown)


nova-gaia:
**Token spend** — 2026-07-10, Opus 4.8 (ultracode / xhigh):
- Orchestration (main loop): ~45k in, ~18k out
- Verification workflow (41 subagents, 441 tool calls): ~2.44M subagent tokens
- **Est. total: ~$40–55**

mbtiongson1 (approved):
lgtm

PR #1167, opened 2026-07-13
feat(cli): add gaia dev rename-named for named-skill renames (#1154)

rico-favor:
Closes #1154.

`gaia dev rename` only handles generic nodes and rejects slashed IDs, so named skills (`contributor/slug`) had no atomic CLI rename path. Adds `gaia dev rename-named <old_id> <new_id>` (new `commands/dev/rename_named.py`) that:

1. Validates both IDs as `contributor/slug` (exactly one slash, both parts non-empty).
2. Moves `registry/named/<contributor>/<slug>.md`.
3. Rewrites frontmatter `id` and `catalogRef` (`contributor-slug`).
4. Updates `suiteComponents` / `suiteRef` / `genericSkillRef` cross-references across all named skills.
5. Appends a `rename` timeline event via `append_skill_event`.

Mirrors the existing `rename` preflight pattern; Verifier-gated like other mutating `gaia dev` verbs.

**Verification:** new `tests/test_rename_named.py` — success path asserts file move, id/catalogRef rewrite, cross-ref rewrites in a sibling skill, and the recorded `rename` timeline event; negative test rejects a slash-less new id (`SystemExit(1)`). Both pass. All imported helpers and `append_skill_event`'s signature verified against source (no invented APIs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mbtiongson1 (changes requested):
Code review (code-only): Well-structured and tests pass (success + negative path). One gap before approval:

The reference rewrite only scans `registry/named/**` (updating `suiteComponents`/`suiteRef`/`genericSkillRef`). The parallel generic `rename` command scans ALL `registry/nodes/**` and updates `derivatives`/`prerequisites` there. Named-skill ids can be referenced from generic nodes, so renaming a named skill risks leaving dangling references in `registry/nodes/**`.

Current data only references named ids via `knownAgents` (agent handles, not skills) and there are no named ids in node `derivatives`/`prerequisites` today, so it is not breaking right now — but for parity with `rename` and future integrity, please also loop `registry/nodes/**/*.json` and update any field equal to / containing `old_id` (at least `derivatives`/`prerequisites`, mirroring rename.py).

Minor: the redundant re-validation of `old_file`/`new_file` existence in `meta_rename_named_command` is fine as defense-in-depth given preflights hard-exit. Once the node-side reference pass is added, I'll approve.

rico-favor:
Author email updated from techteam@favor.church to rico@favor.church via force-pushed rewritten branch history.

PR #1238, opened 2026-07-19
refactor: clean code improvements

raymond-julie:
Hi! 👋

I was looking through the codebase and found a few minor clean code improvements that I thought I'd share.

- **Rule applied:** Magic Numbers, Boolean Naming, DRY

Hope this helps! Thanks for your time and for maintaining this project.


mbtiongson1:
Welcome and thank you for your first contribution! We really appreciate you looking to clean up the code.
Currently, the edits introduced some syntax errors at the top of the file and a regex pattern bug.
Here is how to fix it:
1. Restore the shebang `#!/usr/bin/env python3` on line 1.
2. Define `SHA256_LENGTH = 64` and use an f-string for the regex: `re.compile(fr"^[a-f0-9]{{{SHA256_LENGTH}}}$")`.
3. Separate the function definitions on distinct lines.
4. Copy the updated file to the mirrored path at `.claude/skills/gaia-curate/scripts/validate_discovery_packet.py`.

---

### Why these fixes are necessary:
* **The Syntax Error:** A syntax error at the top of the script blocks the local validator and CI lint steps from running correctly.
* **The Regex Pattern:** Standard raw strings (`r"..."`) do not support variable interpolation. Changing this to an f-string with double curly braces (`fr"..."`) ensures that `SHA256_LENGTH` is evaluated as the integer `64` instead of checking for the literal text `"{SHA256_LENGTH}"`.
* **Mirror Synchronization:** The project maintains identical copies of skill scripts in both `.claude/skills/` and `.agents/skills/`. Keeping them in sync ensures different agent harnesses (like Claude Code, Codex, and Hermes) execute the exact same validation logic.
* **Unused Symbols:** The constants and helpers defined but not used (`MAX_MAPPING_OPTIONS`, `is_valid_decision_value`) should be removed or wired up so the script doesn't carry dead code that clutters future audits.

---

Thank you again for taking the time to contribute to Gaia! Once these adjustments are pushed, the CI pipeline should run green. Let us know if you have any questions or need any help with these steps!

PR #1185, opened 2026-07-15
Yggdrasil II — staging integration (schema + CLI + AOV V4 hero, main merged, PNG history purged)

mbtiongson1:
Fresh aggregate integration PR for `dev/yggdrasil-ii-staging` → `main` (supersedes #1005, which GitHub auto-closed after the history rewrite).

## What's in this branch
- **Yggdrasil II** schema collapse (`{basic,fusion}`, Evidence Floor removed → TM sole gate) and CLI branch-axis (#995/#996).
- **Ascension Overdrive V4** hero — 45 webp assets + V4 CSS/JS runtime; V3 preserved as archive.
- **main fully merged in** — security fixes (RFC #1179 badge, ReDoS #1145, host-validation #1146, token scoping #1144), schema #1151, `scan --dir` #1159, favorchurch #1160, releases → v6.8.8. `main` is an ancestor of this branch.
- **Binary-master history purge** — all ascension/design PNGs + 3 Asset E MP4s expunged from all history via git-filter-repo (webp-only in repo). Fresh clone ~134 MB / 41% lighter. `.gitignore` guard added.

## Notes
- History was rewritten (all branch/tag SHAs changed) — collaborators must re-clone or `git fetch && git reset --hard`.
- Draft: staging still hosts the second half of the sprint. Mark ready when complete.
- All commits authored under the approved mbtiongson1 identity.

Issue #994, opened 2026-07-07
Yggdrasil II · Docs ratification (META, trust-methodology, DESIGN, CONTEXT residual)


mbtiongson1:
**Parent EPIC:** Yggdrasil II (to be created — will link back)
**Ratification handover:** [`founder/handovers/YGGDRASIL_II_RATIFICATION_2026-07-07.md`](../blob/dev/yggdrasil-ii-staging/founder/handovers/YGGDRASIL_II_RATIFICATION_2026-07-07.md)
**Target branch:** `dev/yggdrasil-ii-staging` (NOT main)

## Scope

Lockstep update to canonical docs to reflect Yggdrasil II vocabulary. CONTEXT.md was partially updated in the ratification commit; this PR finishes META.md, trust-methodology.html, DESIGN.md, and any residual CONTEXT drift.

## Files

- `META.md`
  - §1.1 star table — drop Evidence Floor column; add branch rank labels (5★ Ultimate / Unique Ultimate, 6★ Apex / Unique Impossible)
  - §1.2 Skill Types — rewrite from Basic/Extra/Unique/Ultimate to Type axis + Branch axis
  - §4.2 Ultimate & Apex Pathways — split into Suite path (preserved per #935) + Unique path (new gates)
  - §4.3 → split into Suite Apex Gate (unchanged 6 predicates) + §4.4 Unique Impossible Gate (provisional 5 predicates)
  - §8 Implementation Status Tracker — add Yggdrasil II row
- `docs/codex/trust-methodology.html`
  - Evidence Floor section retires; TM Grade Thresholds become the sole gate
  - Apex Gate section splits: Suite Apex + Unique Impossible provisional
  - Terminology switches to "TM Index (2026 Q2)" for external references; keep G7 as internal codename callout only
- `DESIGN.md`
  - Rank sequence: Basic → Awakened → Named → Evolved → Hardened → Ultimate | Unique Ultimate → Apex | Unique Impossible
  - Tier tokens: retire `--tier-unique` / `--tier-ultimate` OR remap to branch-tier tokens (`--branch-unique`, `--branch-suite`)
- `CONTEXT.md` — sanity pass for any missed drift after the ratification commit

## Acceptance

- [ ] Every rank-name reference across the four docs uses Yggdrasil II labels
- [ ] Evidence Floor column removed from META.md §1.1
- [ ] Apex Gate section is explicitly Suite-only
[...]

PR #1240, opened 2026-07-20
Review Design of Staging #1185 - Docs + screenshots + testing

mbtiongson1:
Refs #994


mbtiongson1:
I need to check design branches, I believe that the review here is for some stale plaques, hope we can salvage the previous design session so we dont have to redo. This is getting too complicated, I need to have a simpler branching strat.

mbtiongson1:
Added @INSIGHTS.md for strategic remediation strat

PR #1260, opened 2026-07-23
infra: CI guard against stale hardcoded version stamps

mbtiongson1:
## Version-stamp guard

`window.GAIA_VERSION = "4.3.12"` was hand-typed in source and rotted 6 weeks (repo now 6.8.15) with no guard catching it — lockstep checks only the 4 manifests, and the docs `--check` normalizes the stamp away before comparing.

Adds `scripts/check_version_stamps.py`: fails CI if a literal-semver `window.GAIA_VERSION` stamp appears in `src/**` or `registry/render/**` (dynamic interpolated forms pass). Empty allowlist — the known offenders are FIXED, not whitelisted.

- Fixed `registry/render/gaia.html` (stale 4.3.12 sample -> empty stamp, matching versionless regen output).
- Made `src/gaia_cli/graph.py` stamp dynamic (matches PR 3a).
- Dedicated workflow `version-stamp-guard.yml` (triggers on src/registry-render changes; docs-cohesion only runs at 10+ files, so a dedicated workflow is more reliable for small stamp diffs).

### Scope note
Touches `registry/render/gaia.html` (outside `infra/` allowlist) -> `skip-scope-check` applied (founder standing pre-approval).

### Testing
- Guard exits 0 on the fixed tree; verified it exits 1 + reports file:line on an injected literal stamp.
- Regex discriminates correctly: f-string `{version}`, Jinja `{{ version }}`, graph.py's dynamic form, and empty stamp all pass; only literal semver fails.

Umbrella: #1225

### DIFF
--- a/.github/workflows/version-stamp-guard.yml
+++ b/.github/workflows/version-stamp-guard.yml
@@ -1,0 +1,45 @@
+name: Version Stamp Guard
+
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - 'registry/render/**'
+      - 'scripts/check_version_stamps.py'
+      - '.github/workflows/version-stamp-guard.yml'
+  push:
+    branches:
+      - 'main'
+      - 'dev/**'
+      - 'infra/**'
+    paths:
+      - 'src/**'
+      - 'registry/render/**'
+      - 'scripts/check_version_stamps.py'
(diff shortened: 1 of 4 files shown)


mbtiongson1:
Token spend — 2026-07-23 Opus 4.8 (1m): ~55k in, ~9k out. ~$3

PR #1271, opened 2026-07-25
fix(security): resolve 8 CodeQL high-severity alerts from PR #1185

mbtiongson1:
## Summary

Resolves the **8 high-severity CodeQL alerts** flagged on the Yggdrasil II staging integration PR (#1185, check run 89552403597). All are the JS/Python "unsafe HTML" family; every fix is **behavior-preserving** (identical rendered output / matcher behavior).

## Alerts fixed

| # | File | Rule | Fix |
|---|------|------|-----|
| 1 | `scripts/check_taxonomy_authority.py` | Bad HTML filtering regexp | `</script>` → `</script\s*>` so the guard's inline-`<script>` extractor matches end tags with trailing whitespace (`</script >`) |
| 2–4 | `docs/badges/index.html` | DOM text reinterpreted as HTML | `renderRepoPicker` builds radio labels via the DOM API (`createElement` + `createTextNode`) instead of interpolating user-derived repo names into `innerHTML` — matches the DOM-XSS-safe pattern already used by the sibling `entries.forEach` block |
| 5 | `docs/codex.html` | DOM text reinterpreted as HTML | `addLog()` now escapes by default (`textContent`); the three author-controlled markup lines opt in via `{ html: true }`, and their interpolated `<select>` decision values are run through a new `escLog()` helper |
| 6 | `docs/en/cli-reference.html` | DOM text reinterpreted as HTML | Flag-cell wrap moves existing child nodes into the `.flag-text` span instead of an `innerHTML` read/re-inject round-trip |
| 7 | `docs/js/hoh-modal.js` | DOM text reinterpreted as HTML | Fetched OG SVG is parsed with `DOMParser` (`image/svg+xml`) and adopted via `importNode` instead of `innerHTML` on the fetch response text |
| 8 | `docs/trust/leaderboard/leaderboard.js` | Incomplete string escaping | Inline `onerror` `src` now escapes backslash before quote (`\` then `'`) |

Most inputs are registry-controlled (contributor handles, approved repo names) or constrained `<select>` values, so real-world exploitability was low — but these are legitimate sinks and the fixes are the correct DOM-safe patterns, not suppressions.

## Testing

- `python scripts/check_taxonomy_authority.py` → PASS (0 sites); regex verified against `</script>`, `</script >`, `</script  >`, `<SCRIPT>`.
- `node --check` on `hoh-modal.js`, `leaderboard.js`, and the extracted inline `<script>` blocks of all three HTML files → all parse.
[...]

PR #1272, opened 2026-07-25
infra(security): pre-CodeQL DOM-XSS guard (check_html_sinks.py)

mbtiongson1:
## Summary

Follow-up hardening to #1271. Adds **`scripts/check_html_sinks.py`** — a fast, dataflow-free grep guard for the exact DOM-XSS pattern that produced 6 of the 8 CodeQL alerts on #1185: a runtime value string-built into an HTML sink when a DOM API was available.

It runs **before** CodeQL and fails in-review in <1s, giving authors instant feedback. It does **not** replace CodeQL — CodeQL does real cross-function taint tracking; this catches the common, unambiguous sink line fast. Belt-and-suspenders.

## What it flags

In `docs/**/*.js` and inline `<script>` blocks of `docs/**/*.html`:
- `.innerHTML` / `.outerHTML` assigned a **template literal with `${}`** interpolation
- the same built by **string concatenation** (`'...' + x`)
- `insertAdjacentHTML` / `document.write(ln)` with a dynamic argument

## What it deliberately does NOT flag

- `= ''` / `= ""` (clearing a node)
- a single **static** string literal (constant author-controlled markup)
- **bare-identifier** assignment `el.innerHTML = someVar` — whether `someVar` is tainted is a dataflow question that stays CodeQL's job; flagging it here would be pure false-positive noise
- reads (`const h = el.innerHTML`), comments

## Posture — WARN-ONLY (ratchet)

The repo carries ~14 pre-existing interpolating sinks (mostly `item.innerHTML = \`...${ev.skillName}...\`` in meta-report timelines). Hard-failing all of them today would red-build staging for work outside this PR. So the guard ships **WARN-ONLY** (`HARD_FAIL = False`): it surfaces them as a tracked ratchet and always exits 0. Flip `HARD_FAIL = True` once each interpolating site is converted to a DOM API or annotated.

**Suppression** (for reviewed-safe sinks) — trailing marker with a **required** reason:
```js
el.innerHTML = `<use href="#${ICON}"/>`;  // gaia-html-sink-ok: ICON is a build-time constant
```
A bare marker with no reason is rejected.

## Testing

- `python scripts/check_html_sinks.py --selftest` → **PASS (14 cases)** — safe forms don't flag, dangerous forms do, suppression requires a reason. Wired into CI as a second step.
- `python scripts/check_html_sinks.py` → PASS (warn-only), 20 sinks tracked across the current surface.
- `py_compile` OK; workflow YAML validated.

Structure mirrors the existing `scripts/check_taxonomy_authority.py` guard (same scope globs, inline-`<script>` extraction, `</script\s*>` tolerant regex, output shape).

## Entrypoints

N/A — waived. Tooling/CI only; no user-facing page or surface.

## Notes

- Files: `scripts/check_html_sinks.py`, `.github/workflows/html-sink-guard.yml`, `docs/agents/frontend-known-issues.md` — all within the `infra/` branch allowlist. Self-test folded into the script (`--selftest`) rather than a `tests/` file, keeping the change infra-scoped.
- Base is `dev/yggdrasil-ii-staging`, adjacent to #1271.
- Genuinely new, out-of-scope hardening (not #1185's own remainder) — filed as its own PR per the "no follow-ups for a sprint's own work, but new out-of-scope work is fine" rule.

### DIFF
--- a/.github/workflows/html-sink-guard.yml
+++ b/.github/workflows/html-sink-guard.yml
@@ -1,0 +1,51 @@
+name: HTML Sink Guard (DOM-XSS pre-CodeQL)
+
+on:
+  pull_request:
+    paths:
+      # Trigger on any change to the scanned browser surface or the guard itself
+      - 'docs/**/*.js'
+      - 'docs/**/*.html'
+      - 'scripts/check_html_sinks.py'
+      - '.github/workflows/html-sink-guard.yml'
+  push:
+    branches:
+      - 'main'
+      - 'dev/**'
+      - 'infra/**'
+    paths:
+      - 'docs/**/*.js'
+      - 'docs/**/*.html'
+      - 'scripts/check_html_sinks.py'
+      - '.github/workflows/html-sink-guard.yml'
+  # Allow manual runs for debugging
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  html-sink-guard:
+    name: HTML Sink Guard (pre-CodeQL DOM-XSS)
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Run html-sink guard
+        run: |
+          PYTHONIOENCODING=utf-8 python scripts/check_html_sinks.py
+        # WARN-ONLY posture: always exits 0 today (ratchet, not blocker).
+        # Flip HARD_FAIL=True in the script once interpolating sinks reach zero.
+
+      - name: Guard classifier self-tests
+        run: |
+          PYTHONIOENCODING=utf-8 python scripts/check_html_sinks.py --selftest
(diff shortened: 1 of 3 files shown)


mbtiongson1:
**Token spend (this session, cumulative)**
2026-07-25 Opus 4.8 High: ~150k in, ~24k out. ~$3.60

Scope for this PR: surveyed the ~260-assignment innerHTML surface, authored check_html_sinks.py (warn-only ratchet + --selftest), CI workflow, and frontend-known-issues doc. Adjacent to #1271.

mbtiongson1:
Absorbed into #1274 (cherry-picked as `696407c65`) and closing.

Reviewed with some skepticism before taking it — grep-based guards on `innerHTML` usually need a day-one allowlist, and this repo already carries 358 allowlisted hits on the rank-vocabulary guard, which is how guards decay into noise people stop reading.

This one avoids that. No allowlist: it's **warn-only with a ratchet** (`RESULT: PASS (warn-only) — 20 unsafe sink(s) tracked; flip HARD_FAIL once at zero`) plus a per-line `// gaia-html-sink-ok: <reason>` escape. That's the right shape for a guard that starts non-green.

Useful detail from running it against the tree: of the 20 hits, most are the *same* `item.innerHTML = \`<div style=...\`` line repeated across six generated `docs/meta/reports/*.html` pages, and two more are in `docs/samples/`, which the frontend audit slates for removal. So "20" is really about **3 root causes** — one report-generator template plus a couple of genuine sites. Getting to zero and flipping `HARD_FAIL` is much closer than the number implies. Fix the generator, not the generated HTML.

---

PR #1273, opened 2026-07-25
infra(seo): GA4 analytics injection across all public docs pages

mbtiongson1:
## Summary

- Adds GA4 measurement ID `G-TS7N9T92BM` to all 99 public HTML pages under `docs/` via a new idempotent injector script (`scripts/inject_ga4.py`)
- Strips the dev live-reload script (`localhost:8400/live.js`) that was leaking into production `docs/index.html`
- Skips non-public dirs: `samples/`, `archive/`, `audits/`, `experiments/`

## Entrypoints

No new user-facing pages added — analytics only. N/A for nav/mount registration.

## Files changed

- `scripts/inject_ga4.py` — new idempotent injector; run with `--check` to verify no pages are missing GA4
- `docs/**/*.html` (99 files) — GA4 snippet injected before `</head>`

## Test plan

- [ ] Verify GA4 fires on gaiaskilltree.com after merge (Google Analytics → Realtime view)
- [ ] Confirm `samples/`, `archive/`, `audits/`, `experiments/` pages have no GA4 snippet
- [ ] Re-run `python scripts/inject_ga4.py --check` — should report 0 missing files
- [ ] Confirm `localhost:8400/live.js` no longer appears in `docs/index.html`

Issue #1002, opened 2026-07-07
EPIC · Yggdrasil II — Type + Branch axis split, Ultimate rank rename, Evidence Floor removal


mbtiongson1:
> **Ratification handover:** [`founder/handovers/YGGDRASIL_II_RATIFICATION_2026-07-07.md`](../blob/dev/yggdrasil-ii-staging/founder/handovers/YGGDRASIL_II_RATIFICATION_2026-07-07.md) (already landed on the staging branch)
>
> **Staging branch:** `dev/yggdrasil-ii-staging` — **all sub-issue PRs target this branch, not `main`.** The staging branch is forked from `design/v6.1.1-ascension-overdrive-v2` so the in-flight Ascension redesign moves in lockstep with the taxonomy shift. Staging → main merges **once** at sprint closure.

## The Meta Shift

Yggdrasil II is the second ratified entry in the **Meta Schema RFC series** (Series A — schema-type changes that reshape the skill tree). It splits Gaia's single legacy `type` axis into two orthogonal axes:

- **Type axis** (starless only, structural): collapses `{basic, extra, unique, ultimate}` → `{basic, fusion}`.
- **Branch axis** (named only, progression): `{standard, unique, suite}`, derived at read-time — never declared.
- **Rank renames**: 5★ becomes **Ultimate** (Suite) / **Unique Ultimate** (Unique); 6★ stays **Apex** (Suite) / becomes **Unique Impossible** (Unique). "Ultimate" is now a canonical 5★ rank word (intentional gacha-anchor).
- **Evidence Floor removed.** Trust Magnitude (TM Index (2026 Q2)) is the sole gate.

See the ratification handover for the ten locked decisions (Q1–Q10) from the grill-me interview, the staging-branch protocol, the ubiquitous language glossary, and the migration policy.

## Naming — two RFC series

| Series | Purpose | Latest |
|---|---|---|
| **Meta Schema RFCs** (Series A — Yggdrasil) | Schema-type changes reshaping the tree | **Yggdrasil II** (this EPIC) |
| **TM Index** (Series B — TM cadence) | Trust Magnitude formula / benchmark mechanics | **TM Index (2026 Q2)** ratified (= G7); **TM Index (2026 Q3)** planned (= G8, #1001) |

## Sub-issues (all target `dev/yggdrasil-ii-staging`)

Sequenced Docs → Schema → CLI → Migration → Frontend → CI → Agent Skills; parallelizable where files don't overlap.

[...]

Issue #1000, opened 2026-07-07
Yggdrasil II · Agent skills — refresh curation/audit/fusion prompts for new taxonomy


mbtiongson1:
**Parent EPIC:** Yggdrasil II (to be created)
**Ratification handover:** `founder/handovers/YGGDRASIL_II_RATIFICATION_2026-07-07.md`
**Target branch:** `dev/yggdrasil-ii-staging` (NOT main)

## Scope

Refresh agent-skill prompts and internal curation copy to use Yggdrasil II vocabulary. Low urgency; lands late in the sprint.

## Files

- `.claude/skills/gaia-curate/**` — prompt copy
- `.claude/skills/gaia-curate-chain/**` — prompt copy
- `.claude/skills/gaia-curate-dynamic/**` — prompt copy
- `.claude/skills/gaia-fuse-full-suite/**` — Suite branch semantics; fusion structure vs suiteComponents distinction
- `.claude/skills/gaia-meta-audit/**` — audit rubric under new gates
- `.claude/skills/gaia-audit/**` — same
- `.claude/skills/gaia-draft-curate/**`, `gaia-integrity/**`, `gaia-triage/**` — sanity pass

## Acceptance

- [ ] Every skill prompt references Type + Branch model
- [ ] "Ultimate" used only as 5★ rank word in prompts
- [ ] Fusion structure vs suiteComponents distinction clear where relevant
- [ ] PR base ref is `dev/yggdrasil-ii-staging`

mbtiongson1:
Linked from parent EPIC #1002 (Yggdrasil II).

mbtiongson1:
## RFC #1179 Absorption: Agent skill tree-path updates

Per the RFC #1179 consolidation plan (see #1002 comment), this sub-issue absorbs the agent skill path updates from the skill-trees extraction audit.

### Additional scope (from RFC #1179 Phase 5)

When refreshing agent skills for Yggdrasil II taxonomy vocabulary, also update tree-path references in these skills:

- [ ] `.agents/skills/gaia-trace-timeline/SKILL.md` — explicitly targets `skill-trees/<owner>/skill-tree.json`; update path reference
- [ ] `.agents/skills/gaia-docs-sync/SKILL.md` — note tree decoupling from registry
- [ ] `.agents/skills/gaia-triage/SKILL.md` — indirect timeline event references
- [ ] `.agents/skills/memory-snapshot/SKILL.md` — indirect progression state references
[...]

Issue #995, opened 2026-07-07
Yggdrasil II · Schema — collapse starless type enum to {basic, fusion}


mbtiongson1:
**Parent EPIC:** Yggdrasil II (to be created)
**Ratification handover:** `founder/handovers/YGGDRASIL_II_RATIFICATION_2026-07-07.md`
**Target branch:** `dev/yggdrasil-ii-staging` (NOT main)

## Scope

Collapse the starless `type` enum from 4 values → 2, retire the `type=ultimate → minPrereqs=3` invariant, and mirror changes to the bundled schema snapshot in lockstep.

## Files

- `registry/schema/skill.schema.json`
  - `type` enum: `["basic","extra","unique","ultimate"]` → `["basic","fusion"]`
  - Retire the conditional invariant that required `minPrereqs=3` when `type=ultimate`
  - Preserve `suiteComponents` array field unchanged
- `src/gaia_cli/data/registry/schema/*` — mirror the above (CLAUDE.md schema/ allowlist rule: both schema dirs move in lockstep)
- `registry/schema/meta.json`
  - Rank labels: 5★ split to Ultimate | Unique Ultimate; 6★ split to Apex | Unique Impossible
  - Drop Evidence Floor keys
- `registry/schema/namedSkill.schema.json` — **unchanged** (Option D)

## Acceptance

- [ ] `python scripts/validate.py` meta-sync check passes (both schema dirs identical)
- [ ] `gaia dev validate` passes across the full registry using the new enum
- [ ] `namedSkill.schema.json` has no `type` field added (Option D)
- [ ] PR base ref is `dev/yggdrasil-ii-staging`

## Non-goals

- No CLI arg changes (separate sub-issue, follows this one)
- No migration script (separate sub-issue, follows CLI)

mbtiongson1:
Linked from parent EPIC #1002 (Yggdrasil II).

nova-gaia:
✅ **Done — merged to `dev/yggdrasil-ii-staging` via PR #1173.**

**Delivered:**
- `skill.schema.json` (+ bundled mirror, byte-identical): `type` enum → `[basic, fusion]`; `extra`/`unique`/`ultimate` allOf invariants removed, `basic` (prerequisites maxItems:0) kept.
[...]

PR #1274, opened 2026-07-25
Yggdrasil II — unblock staging: CI greening, performance, agent skills, security

mbtiongson1:
Readies `dev/yggdrasil-ii-staging` for the EPIC #1002 close. **Targets staging, not `main`**, per the EPIC protocol.

Refs #1002 · Closes #1229 · absorbs donor commits from #1271, #1272, #1273

## Why this exists

PR #1185 looked nearly green. It wasn't — it was largely **unvalidated**. Only 9 checks ran on its head; `python-package.yml` (pytest), `validate.yml`, `rank-vocabulary-guard`, `taxonomy-authority-guard`, `version-stamp-guard` and `meta-guard` were all silently skipped, because they are `pull_request` path-filtered and the final push touched only `docs/css/tokens.css`. Those guards fire on `push` to `main` against the whole 1,693-file diff — so every one of them would have gone red *after* merge, not before.

This branch touches `src/` and `tests/`, which forces the full suite to actually run.

## Blockers resolved

| Blocker | Resolution |
|---|---|
| Guard A — 11 hex literals in `tokens.css` | Fixed in `scripts/generateCssTokens.py`, not by hand-editing CSS. The `--ev-type-*` block now emits the same self-documenting `/* var(--x, #hex) */` comment the tier/rank blocks use. Verified with CI's literal grep: 0 hits. |
| 8 CodeQL high-severity alerts | Donor #1271. DOM-safe rewrites (`createElement`/`textContent`, `DOMParser` + `importNode`), not suppressions. |
| Dev live-reload injector in production | `<script src="http://localhost:8400/live.js">` removed from all 6 pages (`index`, `about`, `codex`, `meta`, `privacy`, `starless`). |
| Evidence-Floor dead code and unfailable tests | See below — the most serious finding. |
| `gaia init` / `gaia push` crash | `impl.py:641` and `:2231` indexed `TIER_COLORS['ultimate']` by subscript. `TIER_COLORS` is `{basic, fusion}` post-#995, so both confirmation prompts raised `KeyError`. |
| Dead validators in `scripts/validate.py` | `validate_ultimate()` and `validate_unique_skills()` filtered on retired types, matching zero skills — the pipeline printed `[7/11] Ultimate constraints...` and validated nothing. Removed; steps renumbered to a coherent `[1/9]`–`[9/9]`. |
| `skillBatch.schema.json` admitted retired types | Enum collapsed to `{basic, fusion}` in both canonical and bundled copies (lockstep verified). Writer in `impl.py propose_command()` routed through `taxonomy.isFusion()`. |
| #1000 agent skills | Mirror repaired; retired type literals purged from templates that generated schema-invalid nodes. |

### The Evidence-Floor tests were not red — they were silenced

The prior handover described "7 pre-existing pytest reds" to be modernized later. Those tests no longer existed; they had been deleted in `2b2ee9838` and `74dae4ce7`. What replaced them was worse: `TestGradeTranslation` asserted `_meets_evidence_floor(...) is True`, annotated `(Was: expected False)` — six assertions against a function returning unconditional `True`. Assertions that could not fail.

`EVIDENCE_FLOOR` and `_meets_evidence_floor()` are now deleted from `promotion.py` and `validate.py`. The replacement tests assert the real contract, including that an S-graded self-attestation grades *ungraded* at TM 5 — a case the old floor would have passed at every rank up to 6★.

## Performance

The founder's priority. Measured with headless Chromium against a local server, not inferred.

- **N+1 request storm (#1229).** The homepage fired **312 API requests, 307 of them per-skill JSON files** — three `forEach` loops in `leaderboard.js` each fetching a whole file to read one field (`suiteComponents`, `genericSkillRef`). Those fields were absent from the paginated bulk pages the page *already* loads. They are now carried in the list projection.
- **Unthrottled render loop.** `skill-graph.js` ran `requestAnimationFrame` forever with no visibility gating. Now `IntersectionObserver` + `visibilitychange`: rAF callbacks per 2s measured **15 → 0** when scrolled away or backgrounded.
- **Duplicate JSON fetches.** `cache: 'reload'` bypassed HTTP caching on `gaia.json` (733 KB) on every visit; `skill-graph.js` and `page-ia.js` fetched it independently. Shared `window.GAIA_JSON_CACHE`; `gaia.json` and `named/index.json` both 2→1. A third consumer (`starless.html`) was fetching both at *unversioned* URLs — 4 downloads (~4 MB) → 2.
- **Render-blocking scripts.** Deferred the trailing block (~510 KB). Measured under 4G emulation: `domInteractive` 3643ms → 2302ms (−37%) with no FCP cost. Deferring the *head* scripts was measured and rejected — it delays nav injection, which is FCP, costing ~800ms.

## Not done, deliberately

- **`docs/assets/` unreferenced files** are listed in `founder/handovers/YGGDRASIL_II_DEAD_ASSET_INVENTORY.md`. **Nothing deleted.** The list is basename-heuristic and documents its own limits — one confirmed false positive (`DepartureMono-Regular.woff2`, referenced via `@font-face`) is called out. Founder reviews.
- **Guard A cannot scan `docs/js/skill-graph.js`.** It contains a literal NUL byte (`edge.from + '\0' + edge.to`), so grep classifies it binary and suppresses matches; two banned hexes sit there undetected. Documented, not changed, per founder decision.
- The frontend audit's page-pruning and nav consolidation are a separate sweep.

## Known issue surfaced

Running the full test suite **rewrites Class S artifacts as a side effect** — `docs/graph/gaia.json` shrinks 733,533 → 603,305 bytes and `gaia.gexf` is modified. Committed state here is byte-identical to staging, but any `git add -A` after a test run would ship a truncated graph to the live site. This reproduces independently of this branch's changes.

## Entrypoints

No new user-facing pages or sections. GA4 injection is analytics-only; nav, footer, `window.GAIA_MOUNTS`, and cross-page links are unchanged. N/A.

## Verification

- `python scripts/validate.py` → exit 0, `[1/9]`–`[9/9]`
- Guard A grep → 0 hits
- `check_rank_vocabulary.py` → PASS, 0 hard violations
- `diff -r .claude/skills .agents/skills` → clean but for `fuse/` (vendored third-party with its own LICENSE/CITATION.cff, deliberately not mirrored)
- `pytest tests/` → **1780 passed**, 6 skipped. The single failure is the Class S pollution above, not a code defect — it passes in isolation and on the unmodified staging base.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E9hET1KaRnqcpCz4wR5qid)_


mbtiongson1:
## Baseline comparison — full suite, staging base vs this branch

Ran the complete suite against an unmodified `dev/yggdrasil-ii-staging` worktree to separate pre-existing failures from anything this branch introduces.

| | staging base | this branch |
|---|---|---|
| failed | **8** | **1** |
| passed | 1758 | 1780 |
| skipped | 14 | 6 |

The single remaining failure is `test_world_tree_layout.py::test_world_tree_layout_node_contract`, and it **fails identically on the unmodified base** in a full-suite run. It passes in isolation on both. It is not a defect in this branch.

### Correcting one line in the PR body

The description says that test "passes in isolation and on the unmodified staging base." That's imprecise — it passes *in isolation* on both, and *fails in a full-suite run* on both. The corrected claim is the stronger one: the failure is reproducible on the base and therefore pre-existing.

### Class S pollution is confirmed pre-existing, and it is worse than first measured

After a full suite run the **unmodified base** working tree shows:

```
 M docs/graph/gaia.gexf
 M docs/graph/gaia.json
 M docs/graph/gaia.svg
```

Three Class S artifacts, not two — `gaia.svg` is also rewritten. `gaia.json` drops 733,533 → 603,305 bytes.

This is the live-site substrate. Any contributor who runs the test suite and then stages broadly ships a truncated graph to GitHub Pages — the failure mode behind the 12-hour outage recorded in `EPIC780_OPTION_A_DECISION.md`. Worth its own fix (tests should write to a temp dir), tracked separately from this PR since it predates the branch.

---

## Token spend

2026-07-25 · Opus 5 (orchestrator + 8 delegated agents)

| Workstream | Subagent tokens |
|---|---:|
| Explore — frontend perf audit | 78,106 |
| Explore — CI/test health audit | 127,041 |
| Explore — docs/skills audit (died on session limit; redone inline) | ~50,000 |
| Agent skills mirror + taxonomy prose (#1000) | 148,465 |
| CI greening — Guard A + Evidence-Floor tests | 161,339 |
| Frontend performance | 150,952 |
| validate.py + schema enum | 169,465 |
| API N+1 elimination (#1229) | in flight |

Delegated total ≈ **885k tokens**, plus orchestrator inline. Exact in/out split isn't broken out per agent by the harness, so these are totals rather than the usual in/out format. Est. **~$25–30**.

---

mbtiongson1:
## Final state — N+1 landed, PR ready for review

`claude/epic-1002-api-n1` merged in. **29 commits** against `dev/yggdrasil-ii-staging`.

### The headline number

| Metric | before | after |
|---|---:|---:|
| API requests per homepage load | 312 | **5** |
| per-skill JSON fetches | 307 | **0** |
| per-skill 404s | 78 | **0** |
| total HTTP responses | 361 | **54** |

Cost: **+14.6 KB** across the four paginated pages (63,795 → 78,711 B), well under the 250 KB stop-threshold set for the change.

Render proved **byte-identical**, not merely similar: `.lb-page` innerText string-equal at 3,239 chars; 72 bars total; 40 `#lbSuiteChart` rects with identical ordered `data-id` lists; identical set of 8 dimmed bars; `#lbGenericChart` innerHTML unchanged at 22,373 chars.

The browser's resource-timing buffer was overflowing its 250-entry default before the change and reads 75 after.

### Two things found in passing

- **`fetchUltimateComponents` was already dead code.** `GaiaSemantics.branchOf` reads `entry.branch`, which `boot()` never sets on its row objects, so `ultimates` was always empty and that loop issued zero requests. The 307 came entirely from `detectSuites` and `buildStarlessChart`. It has been rewritten to be fetch-free, but it remains a no-op — **those suite bars do not render today.** Fixing that needs a `branch` field on `allRows`; out of scope here and flagged rather than guessed at.
- **Suite ordering was non-deterministic.** Ties on `trustMagnitude` previously resolved in fetch-completion order. They now resolve in `allRows` order.

### Class S corruption — root-caused, filed as #1275

Not a taxonomy staleness issue, which was the initial hypothesis. Same 243 skills, same 406 edges, **zero** skills differing on any other field. The entire ~130 KB delta is `positions` + `cluster` (243/243 → 0) plus `meta.centroids` and `meta.clusterNames` → null.

Cause: that data derives from `registry/layouts_3d.json`, which is Class P and gitignored, and the build emits a positionless graph instead of failing when it is absent. `tests/test_docs_site.py::test_build_docs_check_message_uses_copyable_python_command` then trips it by running the real build against real repo paths.

This means `tests/world-tree-layout.test.js` was **correctly detecting real corruption** — not a flaky assertion. Committed artifacts here are byte-identical to staging.

### Guards, on the final tip

```
Guard A (hardcoded colors)     PASS — 0 hits
scripts/validate.py            exit 0 — all checks passed
check_rank_vocabulary.py       PASS — 0 hard violations
check_nav_mounts.py (Guard D)  OK
check_html_sinks.py            PASS (warn-only) — 19 tracked
pytest tests/                  1780 passed, 6 skipped
```

Baseline for comparison: the unmodified staging base runs **8 failed / 1758 passed**.

---

PR #1277, opened 2026-07-25
infra(guard-a): harden hex-color guard — NUL fix, binary detector, comment exclusion

mbtiongson1:
## What was broken

Guard A (the CI check that rejects hardcoded hex colour literals in `docs/js/` and `docs/css/`) had two silent gaps:

1. **Silent binary-file skip.** `docs/js/skill-graph.js` carried a literal NUL byte (`'\0'`) used as an edge-key separator. GNU grep treats any file containing a NUL as *binary* and silently skips it, so Guard A never scanned this file at all. Two banned hex literals sat undetected in `//` comments inside it.

2. **No comment exclusion.** Guard A's inline grep could not tell a hex in a `//` comment (documentation) from a hex in live code. Once the NUL was fixed, those comment-form hexes would have tripped the guard for no reason.

3. **No binary-file detector.** Nothing in CI warned that a source file had been silently skipped, so the blind spot could recur unnoticed.

## What each commit does

- **1 — `fix(skill-graph)`**: replaces the literal NUL edge-key separator on `docs/js/skill-graph.js:1713` with the text escape `` (matching `world-tree-layout.js`, which already stores the escape rather than a literal NUL), and strips the hex values from two inline comments (lines 924–925). The file is now plain text — grep no longer skips it.
- **2 — `docs(world-tree-layout)`**: documents the `edgeKey()` NUL-separator convention with a one-line comment warning that the separator must use the `` text escape, never a literal NUL byte (which causes the grep binary-skip Guard A relies on).
- **3 — `infra(guard-a)`**: replaces Guard A's inline grep with `scripts/check_hex_colors.py`. The Python scanner reads every file as text (no binary-skip), strips `//` and `/* */` comments before matching, keeps the allowed `var(--token, #hex)` fallback form, and prints `file:line:hex` on failure. Scope is the live design-system JS/CSS Guard A has always enforced; the generated token-definition file `docs/css/tokens.css` is exempt (it is the one place raw hexes are supposed to live).
- **4 — `infra(guard-a)`**: adds a binary-file NUL detector step that runs *before* the hex scan. It uses `grep -ralP '\x00'` (the `-a` flag forces grep to scan files it would otherwise skip) and fails CI with a clear message listing any `.js`/`.css` under `docs/` that carries a NUL byte.
- **5 — `infra(guard-a)`**: adds `scripts/check_no_binary_sources.py`, a standalone validator (runnable locally: `python scripts/check_no_binary_sources.py`) that checks all `.js`/`.css`/`.html` under `docs/` for NUL bytes and exits non-zero listing offenders.

## Verification

- `python scripts/check_hex_colors.py` exits 0 on the current codebase (comment-form hexes ignored, live-code literals flagged — unit-tested).
- `python scripts/check_no_binary_sources.py` exits 0 (no NUL bytes remain).
- The NUL detector one-liner returns empty on the clean tree and matches a planted NUL file.
- `node --check` passes on both edited JS files; workflow YAML parses.

## Entrypoints

No new user-facing pages. Guard A workflow change only.

## Note on branch scope

This branch is `infra/` but touches `docs/js/*.js` (the NUL fix and the `edgeKey` doc comment — deliverables 1 and 2). The `branch-scope.yml` `infra` allowlist matches `docs/*.html` but not `docs/js/*.js`, so the `skip-scope-check` label is applied for this PR (the JS edits are the core of the sprint and cannot be relocated).

### DIFF
--- a/.github/workflows/docs-cohesion.yml
+++ b/.github/workflows/docs-cohesion.yml
@@ -28,6 +28,29 @@ jobs:
         with:
           fetch-depth: 0
 
+      # ── Guard A — Binary-source detector (runs BEFORE the hex scan) ──────
+      # A NUL byte in a .js/.css source makes GNU grep treat the file as binary
+      # and silently skip it, which is exactly how two banned hexes hid in
+      # docs/js/skill-graph.js undetected. Fail loudly if any source under docs/
+      # carries a NUL byte so a skipped-file gap can never recur silently.
+      - name: Guard A — Detect NUL bytes in JS/CSS sources
+        run: |
+          set +e
+          # -a forces grep to scan files it would otherwise treat as binary and
+          # SKIP — without it this detector would itself miss the NUL-carrying
+          # file, reproducing the very blind spot it exists to close.
+          OFFENDERS=$(grep -ralP '\x00' docs/js/ docs/css/ 2>/dev/null | grep -E '\.(js|css)$')
+          if [ -n "$OFFENDERS" ]; then
+            echo "::error::NUL byte found in JS/CSS source(s) under docs/. A NUL"
+            echo "::error::makes grep treat the file as binary and silently skip"
+            echo "::error::it (Guard A hex scan blind spot). Use a text escape"
+            echo "::error::like '\\u0000' instead of a literal NUL byte."
+            echo "Offending files:"
+            echo "$OFFENDERS"
+            exit 1
+          fi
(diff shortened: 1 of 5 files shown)


mbtiongson1:
2026-07-25 Opus 4.8: ~180k in, ~28k out. ~$8 — Guard A hardening (NUL fix, binary detector, comment-exclusion hex scan, standalone validator).

PR #1280, opened 2026-07-25
fix(docs): guard Class S graph sync against absent layouts_3d.json + fix okf check-mode gap

mbtiongson1:
Fixes #1275

Two pre-existing pipeline bugs on `dev/yggdrasil-ii-staging`, plus a regression re-check of #1274.

## Fix 1 — silent Class S corruption when `registry/layouts_3d.json` is absent

`sync_docs_graph_assets()` loads layout data from `generated-output/layouts.json`, falling back to the committed `registry/layouts_3d.json`. Both are Class P and gitignored, so a fresh clone or CI runner can have neither. `layout_nodes` then stayed `{}` and the function wrote `docs/graph/gaia.json` — the bytes GitHub Pages serves — with no `positions`/`cluster` on any skill and null `meta.centroids`/`meta.clusterNames`. Committing that flattens the live 3D graph and World Tree.

Added a sanity guard in the same shape as `build_badges()`'s catastrophic-drop abort (`scripts/build_docs.py`): if both layout inputs are absent **and** the committed `docs/graph/gaia.json` still carries positions/cluster, raise instead of overwriting. Since this axis is binary rather than proportional, the threshold is simply "committed had positions, regenerated would have none" — no percentage needed. A committed copy that is already positionless is legitimate and passes through untouched.

`build_docs_graph_assets()` now surfaces a non-zero sync as `diff docs/graph/ (sync aborted: rc=N)` in `--check` mode, so the gate fails on the guard instead of swallowing it into a `_run_step` warning.

**Test isolation.** `test_build_docs_check_message_uses_copyable_python_command` asserts on stdout only, but ran the real pipeline against the live tree and rewrote tracked Class S files on every run (`docs/graph/gaia.json` 733533 → 603305 bytes). All build steps are now stubbed via `monkeypatch`, matching the isolation style already used in this file.

## Fix 2 — `build_okf_bundle()` check-mode was a dead check

Check mode returned `False` unconditionally. That silenced one known false positive (`docs/okf/index.json`, whose authoritative writer is `buildSkillsIndex.py`) at the cost of **all** drift detection on the bundle's Markdown outputs.

Check mode now regenerates into a tempdir (`build_okf_bundle.py` takes the output dir as `argv[1]`) and diffs the `.md` outputs against committed `docs/okf/`, printing `diff docs/okf/{relpath}` per mismatch. `index.json` stays excluded with the original rationale kept inline. Write mode is untouched; drift remains warn-only in `main()`.

The new `_diff_md_generated()` helper compares generated → committed only. Committed-only files are ignored deliberately: `build_okf_bundle.py` writes in place and never deletes, so write mode cannot resolve that class of drift, and reporting it would leave `--check` permanently red with no fix command to offer — the trap the original `return False` fell into.

### Pre-existing finding surfaced by this fix

`docs/okf/skills/extra/` and `docs/okf/skills/ultimate/` (≈135 files) are orphans of the Yggdrasil II taxonomy rename — the registry now types skills only `basic` and `fusion`, so the generator no longer emits either directory. They are unreferenced by the regenerated `index.md`. Left in place here: deleting 135 tracked doc files is a deliberate cleanup commit, not a side effect of a check-mode fix. Flagging for a decision.

## Fix 3 — regression re-check of #1274 (no code change)

`gaia init` and `gaia push` were run from this branch's source (`PYTHONPATH=src`) against a scratch git repo, never touching this repo's registry state:

- `gaia init` — clean (the `gaia fetch` HTTP 403 is the sandbox proxy blocking GitHub, handled gracefully).
- `gaia init --force` in a repo with a remote — clean, correctly flips `workspaceMode` to false.
- `gaia scan` — clean.
- `gaia push` — clean; reaches its "No skills to be pushed" guard and exits without a traceback.

The scratch fixture produced no skill detections (no network for `gaia pull`, no real skill corpus), so `push` was not driven through the tree-render path where the KeyError originally lived. Probed that path directly instead: `TIER_COLORS` now contains only `basic` and `fusion`, and every lookup — including `'ultimate'`, the exact key from #1274 — resolves through `.get()` with a fallback. `tier_hex()` and `format_type_colored()` return correctly for `basic`/`extra`/`ultimate`/`fusion`/`unique` and an unknown type. No KeyError; the fix holds.

## Verification

- Guard fires: against a synthetic tree whose committed copy has 243 positioned skills and no layout inputs, it raises and leaves the destination file byte-identical. With a positionless committed copy it writes normally without raising.
- Test isolation: `git status --porcelain` clean before and after; test runs in 0.1s; `docs/graph/gaia.json` stays 733533 bytes.
- OKF check: clean on the unmodified staging tree (no false positives); appending a line to `docs/okf/skills/basic/api-call.md` reports exactly that one path, and nothing else.
- `tests/test_docs_site.py` (3 passed) and `tests/test_seo.py` (12 passed).

## Entrypoints

N/A — no new user-facing page or section. Waived.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MEFBjVQ19RPTzJVCggDBQe

---
_Generated by [Claude Code](https://claude.ai/code/session_01MEFBjVQ19RPTzJVCggDBQe)_

### DIFF
--- a/scripts/build_docs.py
+++ b/scripts/build_docs.py
@@ -323,17 +323,79 @@ def build_okf_bundle(check: bool) -> bool:
         res = subprocess.run([sys.executable, str(script_path)], capture_output=True, text=True)
         return res.returncode == 0
 
-    # In --check mode, do NOT compare build_okf_bundle.py's output against the
-    # committed docs/okf/index.json. Two scripts write that same path in main():
-    # build_okf_bundle (registry/gaia.json) runs first, then build_skills_index
-    # (buildSkillsIndex.py, reads docs/okf/skills/*.md) runs last and wins in
-    # write mode — so the committed index.json is buildSkillsIndex's output.
-    # build_okf_bundle.py only classifies `basic` skills (extra/ultimate come out
-    # empty), so its output can never match the committed file, making this check
-    # a permanent false positive (`diff okf_dir diff_files: ['index.json']`).
-    # build_skills_index (the authoritative writer) has its own --check step, so
-    # genuine docs/okf/index.json drift is still caught there.
-    return False
+    # In --check mode, regenerate into a tempdir and diff only the `.md`
+    # outputs (root index.md, per-type skills/{type}/index.md, per-skill
+    # skills/{type}/{id}.md) against the committed docs/okf/.
+    #
+    # index.json is deliberately EXCLUDED from the comparison. Two scripts
+    # write that same path in main(): build_okf_bundle (registry/gaia.json)
+    # runs first, then build_skills_index (buildSkillsIndex.py, reads
+    # docs/okf/skills/*.md) runs last and wins in write mode — so the committed
+    # index.json is buildSkillsIndex's output. build_okf_bundle.py only
+    # classifies `basic` skills (extra/ultimate come out empty), so its output
+    # can never match the committed file, making that comparison a permanent
+    # false positive (`diff okf_dir diff_files: ['index.json']`).
+    # build_skills_index (the authoritative writer) has its own --check step,
+    # so genuine docs/okf/index.json drift is still caught there.
+    #
+    # Previously this branch returned False unconditionally, which silenced the
+    # index.json false positive at the cost of every other OKF output: real
+    # drift in any generated `.md` went undetected.
+    committed = ROOT / "docs" / "okf"
+    if not committed.is_dir():
+        print("diff docs/okf/ (missing)")
+        return True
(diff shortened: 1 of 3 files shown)


mbtiongson1:
**Token spend (approximate)**

`2026-07-25 Opus 5 Medium: ~90k in, ~12k out. ~$1.50`

Figures are an estimate for this dispatch (2 commits: `a5db4ed`, `029fde7`), not an exact meter reading — flagging that rather than reporting false precision.

---

